PR #2, opened 2020-03-20
Fix patch version when running CMake outside the source tree

dbermond:
Determination of patch version fails when running CMake outside the source tree.

For example, this currently fails to determine the patch version:
```
$ git clone https://github.com/oneapi-src/level-zero.git
$ cmake -B build -S level-zero
$ make -C build
```

To fix this, we must ensure that the `git` command runs in the `${CMAKE_SOURCE_DIR}` directory. This can be done by specifying the source tree with the git `-C` option.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,12 +23,12 @@ else()
 	set(GIT_TAG "v${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}")
 	if(MSVC)
 		execute_process(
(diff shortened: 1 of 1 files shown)


LisannaAtIntel:
We recently pushed https://github.com/oneapi-src/level-zero/commit/1403f9febc122852169675ed5bfd019af47d7cb7 which should resolve this issue. If it does, you can close this pull request.

dbermond:
Yes, commit https://github.com/oneapi-src/level-zero/commit/1403f9febc122852169675ed5bfd019af47d7cb7 fixed this. Thanks.

PR #33, opened 2020-08-12
add newline to error message, add /build to gitignore file

rscohn2:


samples/include/zello_init.h
```diff
@@ -37,7 +37,7 @@ inline bool init_ze( void )
     catch( const ze::exception_t& e )
     {
         std::cout << "Driver not initialized!\n";
-        std::cout << e.what();
+        std::cout << e.what() << "\n";
```
LisannaAtIntel, line 40:
```suggestion
        std::cout << e.what() << std::endl;
```

rscohn2: I changed it to endl. I also put the error info on the same line to make it more obvious that it is the description for the initialization error

PR #38, opened 2020-09-08
Add C++ 14 standard requirement to cmake file

againull:
Level Zero loader sources use make_unique after some recent changes.
This feature is supported only since c++14.

Signed-off-by: Artur Gainullin <artur.gainullin@intel.com>


bmyates:
Thanks, looks good

PR #44, opened 2020-12-22
Fix Fan configuration structure

aquaxp:
From logic point of view and documentation comment 'speedTable' field
should have 'zes_fan_speed_table_t' type instead 'zes_fan_temp_speed_t'.
This field should contain a speed's table that was configured instead
a single entry from speed's table.

### DIFF
--- a/include/zes.py
+++ b/include/zes.py
@@ -841,7 +841,7 @@ class zes_fan_config_t(Structure):
         ("pNext", c_void_p),                                            ## [in][optional] pointer to extension-specific structure
         ("mode", zes_fan_speed_mode_t),                                 ## [in,out] The fan speed mode (fixed, temp-speed table)
         ("speedFixed", zes_fan_speed_t),                                ## [in,out] The current fixed fan speed setting
-        ("speedTable", zes_fan_temp_speed_t)                            ## [out] Array of temperature/fan speed pairs currently configured
+        ("speedTable", zes_fan_speed_table_t)                           ## [out] The currently configured fan speed table
     ]
 
 ###############################################################################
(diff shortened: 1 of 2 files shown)


saik-intel (commented):
Why do you want to change what is not mentioned in spec 
https://spec.oneapi.com/level-zero/latest/sysman/api.html#fan-structures

spec is using  zes_fan_config_t
structzes_fan_config_t
Fan configuration.

Public Members

zes_structure_type_tstype
[in] type of this structure

const void *pNext
[in][optional] pointer to extension-specific structure

zes_fan_speed_mode_tmode
[in,out] The fan speed mode (fixed, temp-speed table)

zes_fan_speed_tspeedFixed
[in,out] The current fixed fan speed setting

zes_fan_temp_speed_tspeedTable
[out] A table containing temperature/speed pairs

aquaxp:
Well, spec is wrong then (and it's anyway generated from the headers).
In current form of spec `speedTable` field is represent "A table containing temperature/speed pairs", but that's obviously not true because used type - `zes_fan_temp_speed_t` is represent only single temperature/speed pair, it's not an array.
To match spec `speedTable` should be at least a literally array - which means that it still need to change it's type from `zes_fan_temp_speed_t` to `zes_fan_temp_speed_t[]`. A proper a data structure that will contain not only array of temperature/speed, but also information how many of them are defined is `zes_fan_speed_table_t`.

`zesFanGetConfig` is intended to return fan configuration and in case if fan is configured for work with table - `speedTable` should contain array of temperature/speed pairs, not a single one.

marcbeuchat:
This is a bug. Thanks for bringing it to our attention. We will fix this in the next release of the spec.

bmyates:
@aquaxp The header files in this repo are generated from the spec.  Once the next release of the spec is published I'll regenerate these headers to pick up this bug fix.

aquaxp:
@bmyates  Should I close this pull-request then and create an issue?

bmyates:
Yes please, thanks!

PR #49, opened 2021-03-29
zes_api: "unkown" strings are "Unknown" in the current implementation

bgoglin:
The current implementation of zesDeviceGetProperties() return "Unknown" instead of "unkown" as stated in the documentation. I assume the documentation is wrong.
I updated the documentation of zes_fabric_port_properties_t as well even if I couldn't find "unkown" nor "Unknown" ever returned there yet.


bgoglin:
Are the specs somewhere else on github? If so, I can look at making a PR there.

PR #50, opened 2021-03-31
Add level-zero.pc

bgoglin:
Compiling against level-zero is currently inconvenient because headers are under include/level_zero instead of include. One has to manually change CPPFLAGS to find them. This PR adds a level-zero.pc on top of libze_loader.pc to make things easy.

Note that the CMake code might be suboptimal but at least it returns what I need for building hwloc with L0 support:
```
$ export PKG_CONFIG_PATH=$PWD/install/lib/pkgconfig/
$ pkg-config level-zero --cflags
-I/home/bgoglin/SOFT/oneapi/level-zero/build/install/include/level_zero -I/home/bgoglin/SOFT/oneapi/level-zero/build/install/include
$ pkg-config level-zero --libs  
-L/home/bgoglin/SOFT/oneapi/level-zero/build/install/lib -lze_loader
```

### DIFF
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -67,4 +67,9 @@ if(UNIX)
         ${CMAKE_CURRENT_BINARY_DIR}/libze_loader.pc
         @ONLY)
     install(FILES "${CMAKE_CURRENT_BINARY_DIR}/libze_loader.pc" DESTINATION "${CMAKE_INSTALL_FULL_LIBDIR}/pkgconfig" COMPONENT level-zero-devel)
+    configure_file(
+        ${CMAKE_CURRENT_SOURCE_DIR}/level-zero.pc.in
+        ${CMAKE_CURRENT_BINARY_DIR}/level-zero.pc
+        @ONLY)
+    install(FILES "${CMAKE_CURRENT_BINARY_DIR}/level-zero.pc" DESTINATION "${CMAKE_INSTALL_FULL_LIBDIR}/pkgconfig" COMPONENT level-zero-devel)
 endif()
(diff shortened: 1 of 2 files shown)


bmyates:
Thanks, looks good

bgoglin:
I was told by some L0 developers that the recommended way is to #include "level_zero/zes_api.h". That's a bit strange because "zes_api.h" does a "#include ze_api.h" instead of "#include level_zero/ze_api.h" but GCC manages to find level_zero/ze_api.h anyway.

If so, I don't know what to do with my PR. libze_loader.pc gives enough cflags to find "level_zero/ze_api.h", but the name "libze_loader.pc" doesn't look like the obvious name of the pkg-config to find ze headers.

bmyates:
Yes, #include `<level_zero/ze_api.h>` is the recommended way.   zes_api.h can do "#include ze_api.h" because they are in the same directory and compilers search current working directory for headers. 

What changes do you think we should make to package config?

bgoglin:
My level-zero.pc doesn't need the level-zero subdirectory in cflags anymore, so it would basically become an empty wrapper that includes libze_loader.pc
We could remove level-zero.pc entirely then (ie revert my PR) but we'd need a clear documentation stating that "libze_loader.pc" is the recommended way to get level_zero cflags despite the strange name of the .pc file.

bgoglin:
I'd like to know if you're going to change or remove my level-zero.pc, my L0-enabled release of hwloc is waiting for this.

bmyates:
If you no longer need level-zero.pc and no one else is using it, then we can remove it.

Would you be able to open a PR with the documentation you are proposing?

bgoglin:
I am note sure where to document these, but if you clarify the way to #include headers for issue #57, you may add a note there saying that libze_loader.pc is the pkg-config file to use for finding those headers. And I gues we could also improve the name and description in libze_loader.pc?

Issue #56, opened 2021-05-10
[Question] Build instructions for Windows? 


jjfumero:
Is there any documentation about how to build level-zero from source on Windows 10? The documentation only specifies for Linux.

Jemale:
Thanks for pointing out this gap @jjfumero, the README should be updated to include instructions for building on Windows in addition to Linux.

The tested/recommended compiler is MSVC including the Spectre-mitigation libraries. CMAKE (tested with version 3.14.0-rc2) can then be used as on Linux:


mkdir build
cd build
cmake ..
cmake --build . --config Release

Jemale:
That configuration is done already in https://github.com/oneapi-src/level-zero/blob/284ccb089184180e34864a9f1e23971d3d736bd8/CMakeLists.txt#L77

The required libraries should be installed with the Visual Studio installer. Under the 'Compilers' section in the 'Individual Components' tab it should look something like this:

![image](https://user-images.githubusercontent.com/1328672/117821155-4676c180-b239-11eb-9dc6-c1c94b214dbc.png)

and under the 'SDKs' section:

![image](https://user-images.githubusercontent.com/1328672/117821391-88076c80-b239-11eb-9fae-08bed94d5bc4.png)

jjfumero:
Thank you @Jemale. It worked. I had to install the missing packages for the Spectre mitigation. 

If you think it is useful, I can open a PR with the documentation for the windows build.

Jemale:
@jjfumero Yes, feel free to do that :)

PR #58, opened 2021-05-21
Add build instructions for Windows

ivorobts:


jjfumero:
Hi @ivorobts , 
   any reason to alter the build instructions from the Linux build? 

This also works on Windows, assuming the  Spectre-mitigation libraries are installed (see https://github.com/oneapi-src/level-zero/issues/56):

```bash
mkdir build
cd build
cmake ..
cmake --build . --config Release
```

Jemale:
@ivorobts Ditto what @jjfumero said. You could list the default CMake invocation and provide setting the nmake generator as a separate example if you want.

ivorobts:
> Hi @ivorobts ,
> any reason to alter the build instructions from the Linux build?
> 
> This also works on Windows, assuming the Spectre-mitigation libraries are installed (see #56):
> 
> ```shell
> mkdir build
> cd build
> cmake ..
> cmake --build . --config Release
> ```

The reason is that I was not able to compile it on Windows without additional actions, e.g. Spectre-mitigation libraries installation. Usage of nmake was much more easier for me. But I agree that we can add instructions for Windows with Visual Studio and as an addition the way I used.

PR #75, opened 2021-11-30
ze_api.h: use ze_bool_t instead of bool for sRGB in ze_srgb_ext_desc_t

bgoglin:
bool doesn't always exist.


bgoglin:
hwloc didn't compile anymore after upgrading level zero packages from 1.4.1 to 1.6.2 using GCC on Debian.

bmyates:
Hey @bgoglin thanks for the PR. We already have this fixed internally and will get it pushed to this repo this week.

bmyates:
This has been fixed here: 7f42f3c6e9599a9bcd995792e2bc42be27ec05ab

Issue #131, opened 2024-02-01
Incorrect libdir in pkg-config files


blue42u:
In some cases, the generated `libze_loader.pc` and `level-zero.pc` produce a `-L` flag (i.e. set `libdir`) to a path that doesn't contain any of the `libze*.so` libraries. On RPM-based platforms this results in link failures if the libraries are not installed to a default path (`cannot find -lze_loader: No such file or directory`), since the path produced refers to `lib` and not the canonical `lib64`.

For example, when installing to `/usr` the `libdir` is missing the architecture triple:
```console
$ cmake -DCMAKE_INSTALL_PREFIX=/usr ..
$ make package
$ dpkg-deb -c level-zero-devel_1.15.8+d_amd64.deb | grep 'libze_[^/]*\.so'
lrwxrwxrwx root/root         0 2024-02-01 16:57 ./usr/lib/x86_64-linux-gnu/libze_loader.so -> libze_loader.so.1
lrwxrwxrwx root/root         0 2024-02-01 16:57 ./usr/lib/x86_64-linux-gnu/libze_tracing_layer.so -> libze_tracing_layer.so.1
lrwxrwxrwx root/root         0 2024-02-01 16:57 ./usr/lib/x86_64-linux-gnu/libze_validation_layer.so -> libze_validation_layer.so.1
$ dpkg-deb --fsys-tarfile level-zero-devel_1.15.8+d_amd64.deb | tar xO ./usr/lib/x86_64-linux-gnu/pkgconfig/libze_loader.pc | head -n3
prefix=/usr
includedir=${prefix}/include
libdir=${prefix}/lib  # Should be ${prefix}/lib/x86_64-linux-gnu
```

When installing to a custom prefix with a custom library subdirectory, the subdirectory is missing from `libdir`:
```console
$ cmake -DCMAKE_INSTALL_PREFIX=/my/install/prefix -DCMAKE_INSTALL_LIBDIR=lib/x86_64-linux-gnu ..
$ make package
$ dpkg-deb -c level-zero-devel_1.15.8+d_amd64.deb | grep 'libze_[^/]*\.so'
lrwxrwxrwx root/root         0 2024-02-01 17:09 ./my/install/prefix/lib/level_zero/libze_loader.so -> libze_loader.so.1
lrwxrwxrwx root/root         0 2024-02-01 17:09 ./my/install/prefix/lib/level_zero/libze_tracing_layer.so -> libze_tracing_layer.so.1
[...]

PR #132, opened 2024-02-01
Fix generated pkg-config files

blue42u:
Fixes https://github.com/oneapi-src/level-zero/issues/131.

The generated files now use the standard paths defined by the [GNUInstallDirs CMake module](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html), which matches the directories the libraries are actually installed to.


blue42u:
Closing in favor of https://github.com/oneapi-src/level-zero/commit/ea655ff5159f93f07c2be920299e64f3c8e52bb3

PR #157, opened 2024-06-17
Change loader log file location

Jemale:


source/loader/ze_loader.cpp
```diff
@@ -261,16 +267,45 @@ namespace loader
...
+                home_dir_string = home_drive + home_path;
+            } else {
+                home_dir_string = ".";
+            }
+        }
+        home_dir = home_dir_string.c_str();
+#else
+        home_dir = getenv("HOME");
+
+        if ((home_dir = getenv("HOME")) == NULL) {
```
lisanna-dettwyler, line 286: Why are you storing it as a C string if it needs to be a std::string at the end?

lisanna-dettwyler (changes requested):
I think this should be reversed, such that the log file is always called ze_loader.log but you can override where it gets put. I can work on this soon.

source/loader/ze_loader.cpp
Jemale: will change

Jemale:
> I think this should be reversed, such that the log file is always called ze_loader.log but you can override where it gets put. I can work on this soon.

I'm not sure I understand, if a user doesn't set `ZEL_LOADER_LOG_FILE` then the log file is `#define LOADER_LOG_FILE_DEFAULT "ze_loader.log"`

lisanna-dettwyler:
What I mean is, eliminate `ZEL_LOADER_LOG_FILE` and introduce `ZEL_LOADER_LOG_DIR`. If it's not set, default to `~/.oneapi_logs`. Rename `LOADER_LOG_FILE_DEFAULT ` to `LOADER_LOG_FILE`. Reasoning is that I just think it's more likely that someone would want to change where the logs get put (like if running in a system where `~` isn't used/writable) than what our specific log file name is.

source/loader/ze_loader.cpp
```diff
@@ -267,37 +267,43 @@ namespace loader
...
+            log_directory = home_dir + "\\" + LOADER_LOG_FILE_DIRECTORY;
 #else
-        home_dir = getenv_string("HOME");
+            home_dir = getenv_string("HOME");
 
-        if (home_dir == "") {
-            auto pwdir = getpwuid(getuid())->pw_dir;
-            home_dir = (pwdir == NULL) ? "." : std::string(pwdir);
-        }
+            if (home_dir == "") {
```
lisanna-dettwyler, line 288: This will never be true because lines 275-281 ensure it's set to something.

source/loader/ze_loader.cpp
```diff
@@ -267,37 +267,43 @@ namespace loader
...
-                home_dir = ".";
+            home_dir = getenv_string("USERPROFILE");
+            if (home_dir == ""){
+                auto home_drive = getenv_string("HOMEDRIVE");
+                auto home_path = getenv_string("HOMEPATH");
+                if ((home_drive != "") && (home_path != "")) {
+                    home_dir = home_drive + home_path;
+                } else {
+                    home_dir = ".";
+                }
```
lisanna-dettwyler, line 282:
```suggestion
                }
```

This allows `home_dir` to still be empty so line 288 can be true. Line 290 should be what sets it to `.`

source/loader/ze_loader.cpp
```diff
@@ -267,37 +267,43 @@ namespace loader
...
+            log_directory = home_dir + "/" + LOADER_LOG_FILE_DIRECTORY;
 #endif
+        }
         auto loader_file = getenv_string("ZEL_LOADER_LOG_FILE");
         if (loader_file.empty()){
-            loader_file = LOADER_LOG_FILE_DEFAULT;
+            auto log_depr_msg = "ZEL_LOADER_LOG_FILE will be deprecated in a future release";
+            loader_file = LOADER_LOG_FILE;
+            std::cout << log_depr_msg << std::endl;
         }
```
lisanna-dettwyler, line 300:
```suggestion
            loader_file = LOADER_LOG_FILE;
        } else {
            auto log_depr_msg = "ZEL_LOADER_LOG_FILE will be deprecated in a future release";
            std::cout << log_depr_msg << std::endl;
        }
```

If it's empty, we don't need to print a deprecation message because it's not being used.

lisanna-dettwyler (changes requested):

source/loader/ze_loader.cpp
Jemale: lines 275-281 are guarded by #ifdef _WIN32, they'll only aply on windows.

lisanna-dettwyler: Ah, didn't see that.

lisanna-dettwyler (dismissed):

third_party/spdlog_headers/spdlog/details/os-inl.h
Jemale, line 1: Was testing and investigated an issue that needed this patch from spdlog upstream: https://github.com/gabime/spdlog/commit/94a8e87c714871dc80478e6636d8fc178f9abd46

lisanna-dettwyler: What were you doing that caused an issue? Setting `ZEL_LOADER_LOG_DIR` to `C:`?

lisanna-dettwyler (approved):

Jemale:
The default location was now being changed to the user home, e.g. C:\Users\jlockett\.oneapi_logs. That was failing due to the same issue in the linked bug

PR #163, opened 2024-06-26
Fix layer init with multiple drivers, perform driver check before init

nrspruit:
-Fix potential race condition when initializing the layers
 with multiple drivers by verifying the usable drivers and layers
 before init of the final ddi tables.


Jemale (approved):
lgtm

lisanna-dettwyler (approved):

PR #164, opened 2024-06-28
Fix layer init with multiple drivers and ensure init for instrumentation support

nrspruit:
-Fix potential race condition when initializing the layers
 with multiple drivers by verifying the usable drivers and layers
 before init of the final ddi tables.

-Document the reasons for requiring ddi table init before driver init
 checks can be done.


lisanna-dettwyler (approved):
There's got to be better way to do this... let's discuss when I'm in Oregon.

PR #165, opened 2024-07-01
Update Changelog for v1.17.17

nrspruit:


lisanna-dettwyler (approved):
Starting at v1.18 I want to switch the versioning to just incrementing the patch number by 1.

PR #167, opened 2024-07-09
Fix DriverGet to handle failed drivers and avoid layer init during checks

nrspruit:
- During init, instrumentation may call driver get before
  uinitialized drivers can be removed. DriverGet has been
  updated to return the driver count only for valid drivers
  and update drivers that are not init to be skipped in
  subsequent DriverGet calls.
- Dont run layer init during checks for driver init to avoid creating
  invalid layer ddi table calls.

### DIFF
--- a/scripts/templates/ldrddi.cpp.mako
+++ b/scripts/templates/ldrddi.cpp.mako
@@ -82,7 +82,13 @@ namespace loader
             uint32_t library_driver_handle_count = 0;
 
             result = drv.dditable.${n}.${th.get_table_name(n, tags, obj)}.${th.make_pfn_name(n, tags, obj)}( &library_driver_handle_count, nullptr );
-            if( ${X}_RESULT_SUCCESS != result ) break;
+            if( ${X}_RESULT_SUCCESS != result ) {
+                // If Get Drivers fails with Uninitialized, then update the driver init status to prevent reporting this driver in the next get call.
+                if (${X}_RESULT_ERROR_UNINITIALIZED == result) {
+                    drv.initStatus = result;
+                }
+                continue;
+            }
 
             if( nullptr != ${obj['params'][1]['name']} && *${obj['params'][0]['name']} !=0)
             {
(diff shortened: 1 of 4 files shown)


scripts/templates/ldrddi.cpp.mako
```diff
@@ -82,7 +82,13 @@ namespace loader
             uint32_t library_driver_handle_count = 0;
 
             result = drv.dditable.${n}.${th.get_table_name(n, tags, obj)}.${th.make_pfn_name(n, tags, obj)}( &library_driver_handle_count, nullptr );
-            if( ${X}_RESULT_SUCCESS != result ) break;
+            if( ${X}_RESULT_SUCCESS != result ) {
```
Jemale, line 85: Will it never be the case that a driver init status could change from Uninitialized after the first call?

source/loader/ze_loader.cpp
```diff
@@ -200,46 +200,6 @@ namespace loader
             return ZE_RESULT_ERROR_UNINITIALIZED;
         }
 
-        if(nullptr != validationLayer) {
```
Jemale, line 203: Are these sections not needed at all anymore?

nrspruit: These sections were being extra careful by checking that not only did the call to zeInit() pass going to the driver, but thru the validation and tracing layers if they exist.

This check should not be required. The only time this check caught anything is if one had an invalid set of layers installed, but then the error would occur at zeInit() time frame instead of during the init phase.

The addition of the handling for at least one driver success in zeDriverGet removes the requirement for these checks.

scripts/templates/ldrddi.cpp.mako
nrspruit: The init status will be updated on the call to zeInit for that driver, but that will be before the call to driverget.

https://github.com/oneapi-src/level-zero/blob/master/source/loader/ze_ldrddi.cpp#L52

The only time the init status may change after a call is if we are in instrumentation callback, but then the instrumentation excluded driver types would be the one that failed.

Therefore, the call to driver get from the instrumentation callback will only get the drivers that the init flag allowed, but the later call to init that allows all drivers would get the full set of drivers because the init flag was updated.

source/loader/ze_loader.cpp
Jemale: sounds good

Jemale (dismissed):

scripts/templates/ldrddi.cpp.mako
```diff
@@ -82,7 +82,13 @@ namespace loader
...
+                // If Get Drivers fails with Uninitialized, then update the driver init status to prevent reporting this driver in the next get call.
+                if (${X}_RESULT_ERROR_UNINITIALIZED == result) {
+                    drv.initStatus = result;
+                }
+                continue;
```
nrspruit, line 90: I realized that we actually don't want to "break", but continue to go thru the drivers now that the DriverGet is allowed to get a failure.

Jemale (approved):

bmyates (approved):
While I don't fully understand all edge cases here, in general looks good.  Do we have test cases for this?

nrspruit:
> While I don't fully understand all edge cases here, in general looks good. Do we have test cases for this?

test_pin in the CTS with intercept layer enabled.

PR #177, opened 2024-08-07
Fix first zeinit to allow for layer checks

nrspruit:
- Use ddi table init of zeinit for the first call to zeInit to enable intercept in layers.

### DIFF
--- a/source/loader/ze_loader.cpp
+++ b/source/loader/ze_loader.cpp
@@ -124,7 +124,7 @@ namespace loader
         }
     }
 
-    ze_result_t context_t::check_drivers(ze_init_flags_t flags) {
+    ze_result_t context_t::check_drivers(ze_init_flags_t flags, ze_global_dditable_t *globalInitStored) {
         if (debugTraceEnabled) {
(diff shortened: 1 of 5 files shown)


source/loader/ze_loader.cpp
```diff
@@ -209,7 +209,8 @@ namespace loader
             return ZE_RESULT_ERROR_UNINITIALIZED;
         }
 
-        ze_result_t res = pfnInit(flags);
+        // Use the previously init ddi table pointer to zeInit to allow for intercept of the zeInit calls
+        ze_result_t res = globalInitStored->pfnInit(flags);
```
Jemale, line 213: is `globalInitStored` always guaanteed to be valid/non-null?

Jemale (dismissed):
looks fine to me, just had a general question

source/loader/ze_loader.cpp
nrspruit: Good point, globalInitStored is guranteed to be valid otherwise the GetProcTable api would have failed and we would not get to this code. But if we have a malformed driver the pfnInit could be nullptr, I fixed the previous check to ensure both the init from the driver and the table are not null.

Jemale (approved):

PR #185, opened 2024-08-27
Fix handle translation when ddi table fallback

nrspruit:
- When enough drivers are removed that ddi table fallback is completed, then intercept_enabled needs to be set to false such that handle translation is done correctly.


Jemale (dismissed):

lisanna-dettwyler (dismissed):

Jemale (dismissed):
Looks fine to me

lisanna-dettwyler (approved):

PR #192, opened 2024-09-03
Prevent reinit when zes/ze handles have already been retrieved

nrspruit:
- Given ze or zes driver get has been called, then one cannot allow for the ddi table to be reinit to avoid breaking the user's existing handles.


Jemale (approved):
why is the program instrumentation check not needed anymore?

nrspruit:
> why is the program instrumentation check not needed anymore?

It is replaced with the zeInuse check.

The problem with instrumentation was the calls to L0 apis during zeinit().

With the new zeInuse, calls to zeDriverGet while in the driver_check will show up and when ze has already started to be used.

so, this fix includes the support for handling program instrumentation cases.

PR #193, opened 2024-09-04
Fix SysMan driver pointers given no sysman env and no zesInit

nrspruit:
- To gracefully allow calls to zes* apis without setting the env nor calling zesinit, the drivers need to be defaulted to the zeDrivers and have the zes ddi tables always updated unless ze or zes handles have already been retrieved.

### DIFF
--- a/source/lib/ze_lib.cpp
+++ b/source/lib/ze_lib.cpp
@@ -74,6 +74,11 @@ namespace ze_lib
             ze_lib::context->zesDdiTable.exchange(&ze_lib::context->initialzesDdiTable);
         }
 
+        // Given zesInit, then zesDrivers needs to be used as the sysmanInstanceDrivers;
+        if (sysmanOnly) {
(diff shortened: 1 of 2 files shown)


source/lib/ze_lib.cpp
```diff
@@ -74,6 +74,11 @@ namespace ze_lib
             ze_lib::context->zesDdiTable.exchange(&ze_lib::context->initialzesDdiTable);
         }
 
+        // Given zesInit, then zesDrivers needs to be used as the sysmanInstanceDrivers;
+        if (sysmanOnly) {
+            loader::context->sysmanInstanceDrivers = &loader::context->zesDrivers;
```
lisanna-dettwyler, line 79: Just verifying, this runs *after* the change in ze_loader.cpp?

lisanna-dettwyler (approved):

source/lib/ze_lib.cpp
nrspruit: yes

PR #200, opened 2024-09-10
v1.11 Spec Changes for the L0 Loader

nrspruit:
- Generated code changes for the Loader supporting L0 v1.11.0
- Added new metric handle generators
- Add support for zeInitDrivers


nrspruit:
Hello @lisanna-dettwyler / @rwmcguir / @Jemale ,

This is the first draft of the v1.10 support, I still need to write tests for zeInitDrivers and expand the null driver for that testing, but this is the first working draft.

source/lib/ze_lib.cpp
```diff
@@ -114,7 +114,7 @@ namespace ze_lib
             // Check which drivers support the ze_driver_flag_t specified
             // No need to check if only initializing sysman
             bool requireDdiReinit = false;
-            result = zelLoaderDriverCheck(flags, &ze_lib::context->initialzeDdiTable.Global, &ze_lib::context->initialzesDdiTable.Global, &requireDdiReinit, sysmanOnly);
+            result = zelLoaderDriverCheck(flags, desc, &ze_lib::context->initialzeDdiTable.Global, &ze_lib::context->initialzesDdiTable.Global, &requireDdiReinit, sysmanOnly);
```
lisanna-dettwyler, line 117: am I just now noticing that a lot of this function assumes `! DYNAMIC_LOAD_LOADER`?

nrspruit: yes, if we enable that option there will need to be a refactor.

nrspruit:
@lisanna-dettwyler we need to bump the minor version of the Loader in this PR as well ie to 1.18 , I realized that is missing in this PR currently.

lisanna-dettwyler (approved):

PR #213, opened 2024-10-08
Fixed EXP handling and removed invalid ddi table

nrspruit:


nrspruit:
abi check is expected to fail because we removed the invlaid non-exp ddi table.

bmyates (approved):

rwmcguir (approved):

matcabral (approved):
metrics function pointers look good

PR #216, opened 2024-10-14
Fix EXP Headers for ddi tables

nrspruit:


include/zet_ddi.h
```diff
@@ -737,11 +737,11 @@ typedef ze_result_t (ZE_APICALL *zet_pfnMetricGroupDestroyExp_t)(
 /// @brief Table of MetricGroupExp functions pointers
 typedef struct _zet_metric_group_exp_dditable_t
 {
-    zet_pfnMetricGroupCreateExp_t                               pfnCreateExp;
     zet_pfnMetricGroupCalculateMultipleMetricValuesExp_t        pfnCalculateMultipleMetricValuesExp;
     zet_pfnMetricGroupGetGlobalTimestampsExp_t                  pfnGetGlobalTimestampsExp;
     zet_pfnMetricGroupGetExportDataExp_t                        pfnGetExportDataExp;
     zet_pfnMetricGroupCalculateMetricExportDataExp_t            pfnCalculateMetricExportDataExp;
+    zet_pfnMetricGroupCreateExp_t                               pfnCreateExp;
```
matcabral, line 744: shouldn't this go at the end  of the struct?

nrspruit: that would break backwards compatibility, the version from v1.9.3 had this symbol in this place. I am simply returning it to the right place.

include/ze_ddi.h
```diff
@@ -1200,14 +1200,14 @@ typedef ze_result_t (ZE_APICALL *ze_pfnCommandListImmediateAppendCommandListsExp
...
-    ze_pfnCommandListUpdateMutableCommandKernelsExp_t           pfnUpdateMutableCommandKernelsExp;
     ze_pfnCommandListCreateCloneExp_t                           pfnCreateCloneExp;
+    ze_pfnCommandListImmediateAppendCommandListsExp_t           pfnImmediateAppendCommandListsExp;
     ze_pfnCommandListGetNextCommandIdExp_t                      pfnGetNextCommandIdExp;
+    ze_pfnCommandListUpdateMutableCommandsExp_t                 pfnUpdateMutableCommandsExp;
+    ze_pfnCommandListUpdateMutableCommandSignalEventExp_t       pfnUpdateMutableCommandSignalEventExp;
     ze_pfnCommandListUpdateMutableCommandWaitEventsExp_t        pfnUpdateMutableCommandWaitEventsExp;
-    ze_pfnCommandListImmediateAppendCommandListsExp_t           pfnImmediateAppendCommandListsExp;
+    ze_pfnCommandListGetNextCommandIdWithKernelsExp_t           pfnGetNextCommandIdWithKernelsExp;
+    ze_pfnCommandListUpdateMutableCommandKernelsExp_t           pfnUpdateMutableCommandKernelsExp;
```
bmyates, line 1210: thisn is not the same ordering i get with update to spec

include/zet_ddi.h
nrspruit: based on testing with https://github.com/oneapi-src/level-zero/pull/217 this may not be necessary in this case.

PR #220, opened 2024-10-15
Update workflows to ABI check with last released tag

nrspruit:
Expanded ABI checker workflow to verify both the previous commit and the previous release tag.


nrspruit:
@lisanna-dettwyler / @Jemale  / @rwmcguir  please let me know if we can add this to the CI, this will catch more ABI issues related to breaking backwards compatibility with previous releases.

lisanna-dettwyler (approved):
I want to condense this into being a matrix job, but that refactor can happen later

PR #223, opened 2024-10-23
Fix Backwards compatibility in ddi structs for EXP features

nrspruit:
- Fix incorrect reordering from 1.9.x of the EXP function pointers in ddi tables.

### DIFF
--- a/include/ze_ddi.h
+++ b/include/ze_ddi.h
@@ -5,7 +5,7 @@
  * SPDX-License-Identifier: MIT
  *
  * @file ze_ddi.h
- * @version v1.11-r1.11.3
+ * @version v1.11-r1.11.4
  *
  */
 #ifndef _ZE_DDI_H
@@ -1135,23 +1135,6 @@ typedef ze_result_t (ZE_APICALL *ze_pfnCommandListCreateCloneExp_t)(
     );
 
 ///////////////////////////////////////////////////////////////////////////////
-/// @brief Function-pointer for zeCommandListGetNextCommandIdExp 
-typedef ze_result_t (ZE_APICALL *ze_pfnCommandListGetNextCommandIdExp_t)(
-    ze_command_list_handle_t,
-    const ze_mutable_command_id_exp_desc_t*,
-    uint64_t*
-    );
-
-///////////////////////////////////////////////////////////////////////////////
-/// @brief Function-pointer for zeCommandListUpdateMutableCommandWaitEventsExp 
-typedef ze_result_t (ZE_APICALL *ze_pfnCommandListUpdateMutableCommandWaitEventsExp_t)(
-    ze_command_list_handle_t,
-    uint64_t,
-    uint32_t,
-    ze_event_handle_t*
-    );
-
-///////////////////////////////////////////////////////////////////////////////
 /// @brief Function-pointer for zeCommandListImmediateAppendCommandListsExp 
 typedef ze_result_t (ZE_APICALL *ze_pfnCommandListImmediateAppendCommandListsExp_t)(
     ze_command_list_handle_t,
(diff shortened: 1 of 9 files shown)


include/zet_ddi.h
```diff
@@ -565,9 +565,9 @@ typedef ze_result_t (ZE_APICALL *zet_pfnMetricCreateFromProgrammableExp_t)(
 /// @brief Table of MetricExp functions pointers
 typedef struct _zet_metric_exp_dditable_t
 {
+    zet_pfnMetricCreateFromProgrammableExp_t                    pfnCreateFromProgrammableExp;
     zet_pfnMetricDestroyExp_t                                   pfnDestroyExp;
     zet_pfnMetricCreateFromProgrammableExp2_t                   pfnCreateFromProgrammableExp2;
-    zet_pfnMetricCreateFromProgrammableExp_t                    pfnCreateFromProgrammableExp;
```
bmyates, line 570: The spec shows this API as being added in 1.11.   The change was made here  https://github.com/oneapi-src/level-zero-spec/commit/bd119b848b666de6494053e61c425d249e751746

This is the reason it is at end of table.

nrspruit: that should not have changed, the old api should have remained 1.9, that is the bug.

include/ze_ddi.h
```diff
@@ -1201,12 +1201,12 @@ typedef ze_result_t (ZE_APICALL *ze_pfnCommandListUpdateMutableCommandKernelsExp
...
 {
     ze_pfnCommandListCreateCloneExp_t                           pfnCreateCloneExp;
-    ze_pfnCommandListGetNextCommandIdExp_t                      pfnGetNextCommandIdExp;
-    ze_pfnCommandListUpdateMutableCommandWaitEventsExp_t        pfnUpdateMutableCommandWaitEventsExp;
     ze_pfnCommandListImmediateAppendCommandListsExp_t           pfnImmediateAppendCommandListsExp;
-    ze_pfnCommandListGetNextCommandIdWithKernelsExp_t           pfnGetNextCommandIdWithKernelsExp;
+    ze_pfnCommandListGetNextCommandIdExp_t                      pfnGetNextCommandIdExp;
     ze_pfnCommandListUpdateMutableCommandsExp_t                 pfnUpdateMutableCommandsExp;
     ze_pfnCommandListUpdateMutableCommandSignalEventExp_t       pfnUpdateMutableCommandSignalEventExp;
+    ze_pfnCommandListUpdateMutableCommandWaitEventsExp_t        pfnUpdateMutableCommandWaitEventsExp;
```
bmyates, line 1208: `UpdateMutableCommandWaitEventsExp`  was added in version 1.9 

`UpdateMutableCommandSignalEventExp`  was added in 1.10

Yet here we have 1.9 API listed after 1.10 API.  That does not see right.

nrspruit:
The issue was that the .yaml files were updated such that they indicated that features that were added in 1.9 were "new" in 1.10/1.11, which is not true, they had to be still 1.9 versioned items otherwise the spec scripts cannot order them properly.

bmyates:
> The issue was that the .yaml files were updated such that they indicated that features that were added in 1.9 were "new" in 1.10/1.11, which is not true, they had to be still 1.9 versioned items otherwise the spec scripts cannot order them properly.

Got it.  I'd recommend fixing yaml files and verifying the generated headers match what you have here.  Otherwise this issue is gonna keep coming up and biting us every release

bmyates (dismissed):

aravindksg (approved):

matcabral (approved):
tool file looks ok

lisanna-dettwyler (changes requested):
This should be in draft state until the spec changes are merged and released/versioned

nrspruit:
> This should be in draft state until the spec changes are merged and released/versioned

Changes have been updated with the latest code in https://github.com/oneapi-src/level-zero-spec/releases/tag/v1.11.1

bmyates (dismissed):

aravindksg (dismissed):

include/ze_ddi.h
```diff
@@ -5,7 +5,7 @@
  * SPDX-License-Identifier: MIT
  *
  * @file ze_ddi.h
- * @version v1.11-r1.11.3
+ * @version v1.11-r1.11.1
```
lisanna-dettwyler, line 8: The version number is going backwards?!

nrspruit: The revision comment has been fixed now it matches the actual tagged revision in the spec repo vs the "commit count" the generated headers were adding, ie :
https://github.com/oneapi-src/level-zero-spec/releases

nrspruit: note, the 1.11.1 is generated by the scripts as of the latest master branch in https://github.com/oneapi-src/level-zero-spec so this is not something changed when I posted the patch.

lisanna-dettwyler: Seems like we should ask Pradeep to fix the version to be 1.11.4 so we're not rewriting history.

nrspruit: updated with the 1.11.4 version

PR #224, opened 2024-10-25
Static L0 Loader Support

nrspruit:
Static Loader Support.

This support has a few conditions:

1. zes/ze driver split is not supported without this version of the loader so backwards compatibility for different lists of drivers for ze/zes init is not supported if using the static lib due to needing to modify the loader's context. zes/ze driver list will be the same.
2. As of this version of the loader, all symbols read between the static loader and the dynamic loader must remain ABI compatible ie all apis exported in source/loader/ze_loader_api.h now must be abi compatible.
3. The oldest dynamic loader supported is v1.15.17

### DIFF
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -12,11 +12,22 @@ configure_file(
     @ONLY)
 
 include(GNUInstallDirs)
-add_library(${TARGET_LOADER_NAME}
-    SHARED
-    ""
-    ${CMAKE_CURRENT_BINARY_DIR}/ZeLoaderVersion.rc
-)
+if (DYNAMIC_LOAD_LOADER)
+    message(STATUS "Building loader as static library")
+    add_library(${TARGET_LOADER_NAME}
+        STATIC
+        ""
+        ${CMAKE_CURRENT_BINARY_DIR}/ZeLoaderVersion.rc
+    )
+    add_definitions(-DDYNAMIC_LOAD_LOADER="1")
+else()
+    message(STATUS "Building loader as dynamic library")
+    add_library(${TARGET_LOADER_NAME}
+        SHARED
+        ""
+        ${CMAKE_CURRENT_BINARY_DIR}/ZeLoaderVersion.rc
+    )
+endif()
 
 add_subdirectory(lib)
 add_subdirectory(loader)
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -12,7 +12,7 @@ configure_file(
     @ONLY)
 
 include(GNUInstallDirs)
-if (DYNAMIC_LOAD_LOADER)
+if (BUILD_STATIC)
     message(STATUS "Building loader as static library")
     add_library(${TARGET_LOADER_NAME}
         STATIC
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -30,8 +30,10 @@ else()
 endif()
 
 add_subdirectory(lib)
-add_subdirectory(loader)
-add_subdirectory(layers)
+if (NOT BUILD_STATIC)
+    add_subdirectory(loader)
+    add_subdirectory(layers)
+endif()
 add_subdirectory(drivers)
 add_subdirectory(utils)
 
--- a/source/lib/ze_lib.cpp
+++ b/source/lib/ze_lib.cpp
@@ -69,16 +69,32 @@ namespace ze_lib
         typedef HMODULE (ZE_APICALL *getTracing_t)();
         auto getTracing = reinterpret_cast<getTracing_t>(
             GET_FUNCTION_PTR(loader, "zeLoaderGetTracingHandle") );
+        if (getTracing == nullptr) {
+            printf("missing getTracing\n");
+            return ZE_RESULT_ERROR_UNINITIALIZED;
+        }
         tracing_lib = getTracing();
         typedef ze_result_t (ZE_APICALL *zelLoaderDriverCheck_t)(ze_init_flags_t flags, ze_init_driver_type_desc_t* desc, ze_global_dditable_t *globalInitStored, zes_global_dditable_t *sysmanGlobalInitStored, bool *requireDdiReinit, bool sysmanOnly);
         auto loaderDriverCheck = reinterpret_cast<zelLoaderDriverCheck_t>(
                 GET_FUNCTION_PTR(loader, "zelLoaderDriverCheck") );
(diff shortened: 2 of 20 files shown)


nrspruit:
example output:
scss_dev@nuc612dut2:~/level-zero-tests/build$ ./output/test_init_sysman
Context created static
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from SysmanInitTests
[ RUN      ] SysmanInitTests.GivenCoreNotInitializedWhenSysmanInitializedThenzesDriverGetWorks
Context created context lib dynamic
Context created dynamic
loader created context lib dynamic
calling loader init static
ZE_LOADER_DEBUG_TRACE:Using Loader Library Path: 
ZE_LOADER_DEBUG_TRACE:Loading Driver libze_intel_gpu.so.1
ZE_LOADER_DEBUG_TRACE:Loading Driver libze_intel_gpu_legacy1.so.1
ZE_LOADER_DEBUG_TRACE:Load Library of libze_intel_gpu_legacy1.so.1 failed with libze_intel_gpu_legacy1.so.1: cannot open shared object file: No such file or directory
ZE_LOADER_DEBUG_TRACE:Loading Driver libze_intel_vpu.so.1
ZE_LOADER_DEBUG_TRACE:Load Library of libze_intel_vpu.so.1 failed with libze_intel_vpu.so.1: cannot open shared object file: No such file or directory
ZE_LOADER_DEBUG_TRACE:Loading Driver libze_intel_npu.so.1
ZE_LOADER_DEBUG_TRACE:Load Library of libze_intel_npu.so.1 failed with libze_intel_npu.so.1: cannot open shared object file: No such file or directory
ZE_LOADER_DEBUG_TRACE:Tracing Layer Library Path: libze_tracing_layer.so.1
done loading functions
zelLoaderGetContext
calling static loader ddi init
ZE_LOADER_DEBUG_TRACE:check_drivers(flags=0(ZE_INIT_ALL_DRIVER_TYPES_ENABLED))
ZE_LOADER_DEBUG_TRACE:init driver libze_intel_gpu.so.1 zesInit(0(ZE_INIT_ALL_DRIVER_TYPES_ENABLED)) returning ZE_RESULT_SUCCESS
[       OK ] SysmanInitTests.GivenCoreNotInitializedWhenSysmanInitializedThenzesDriverGetWorks (29 ms)
[----------] 1 test from SysmanInitTests (29 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (29 ms total)
[  PASSED  ] 1 test.
loader destroyed context lib dynamic
Context destroyed context lib dynamic
scss_dev@nuc612dut2:~/level-zero-tests/build$ ldd ./output/test_init_sysman
        linux-vdso.so.1 (0x00007ffc7d5d8000)
        libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f1d4a200000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f1d4a519000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f1d4a922000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f1d49e00000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f1d4a958000)

source/CMakeLists.txt
```diff
@@ -12,11 +12,22 @@ configure_file(
     @ONLY)
 
 include(GNUInstallDirs)
-add_library(${TARGET_LOADER_NAME}
-    SHARED
-    ""
-    ${CMAKE_CURRENT_BINARY_DIR}/ZeLoaderVersion.rc
-)
+if (DYNAMIC_LOAD_LOADER)
```
lisanna-dettwyler, line 15: Recommend the top level cmake option be named after the result rather than the behavior, like `BUILD_STATIC`

source/lib/ze_lib.cpp
```diff
@@ -46,19 +53,48 @@ namespace ze_lib
...
+                GET_FUNCTION_PTR(loader, "zelLoaderDriverCheck") );
+        if (loaderDriverCheck == nullptr) {
+            printf("missing loaderDriverCheck\n");
+            return ZE_RESULT_ERROR_UNINITIALIZED;
         }
+        typedef ze_result_t (ZE_APICALL *zelLoaderTracingLayerInit_t)(std::atomic<ze_dditable_t *> &zeDdiTable);
+        auto loaderTracingLayerInit = reinterpret_cast<zelLoaderTracingLayerInit_t>(
+                GET_FUNCTION_PTR(loader, "zelLoaderTracingLayerInit") );
+        if (loaderTracingLayerInit == nullptr) {
+            printf("missing loaderTracingLayerInit\n");
```
lisanna-dettwyler, line 88: I think these should be going to stderr

lisanna-dettwyler (commented):
Just a couple small comments from looking at it briefly

source/lib/ze_lib.cpp
nrspruit: agreed, I was going to convert the printfs to debug trace logs for the ones we determine useful.

source/CMakeLists.txt
nrspruit: good point, that would differentiate.

nrspruit: Fixed in the latest patches.

source/lib/ze_lib.cpp
nrspruit: Fixed in the latest patches.

source/lib/ze_lib.h
```diff
@@ -34,22 +36,130 @@ namespace ze_lib
         context_t();
         ~context_t();
 
+    //////////////////////////////////////////////////////////////////////////
+    std::string to_string(const ze_result_t result) {
```
rwmcguir, line 40: I really feel like this should be coming in the form of a header being generated at build time, else we have a maintanance nightmare..  something clever with sed or perhaps something formed by Mako.

rwmcguir, line 40: This work, but have you considered using stringify w/preprocessor macros? 
Which may also work to not miss future ZE_RESULT additions

Trying to avoid a future maintenance issue

rwmcguir:
Looks good to me, as I have seen the testing output. Debating on if we need to increment minor here as part of this commit since it is a decent size feature add-on.

rwmcguir (approved):

PR #226, opened 2024-10-30
Events deadlock

bibrak:
Creates an internal DAG of the events at runtime. From that DAG maintains a topological sort. If there is a cycle in the DAG, adding that dependency in the topological sort will return error hence detecting the deadlock.

### DIFF
--- a/source/layers/validation/README.md
+++ b/source/layers/validation/README.md
@@ -18,6 +18,7 @@ By default, no validation modes will be enabled. The individual validation modes
 
 - `ZE_ENABLE_PARAMETER_VALIDATION`
 - `ZE_ENABLE_HANDLE_LIFETIME`
+- `ZEL_ENABLE_EVENTS_CHECKER`
 - `ZE_ENABLE_MEMORY_TRACKER` (Not yet Implemeneted)
 - `ZE_ENABLE_THREADING_VALIDATION` (Not yet Implemeneted)
 
@@ -48,6 +49,11 @@ This mode maintains an internal mapping of each handle type to a state structure
 - Additional per handle state checks added as needed
     - Example - Check ze_cmdlist_handle_t open or closed
 
+### `ZEL_ENABLE_EVENTS_CHECKER`
+
+The Events Checker validates usage of events. 
+- It is designed to detect potential deadlocks that might occur due to improper event usage in the Level Zero API. It prints out warning messages for user when it detects a potential deadlock.
+- In some cases it may also detect whether an event is being used more than once without being reset. Consider a case in which a single event is signaled from twice.
 
 ### `ZE_ENABLE_THREADING_VALIDATION` (Not yet Implemeneted)
 
--- a/third_party/xla/graphcycles.h
+++ b/third_party/xla/graphcycles.h
@@ -1,0 +1,167 @@
+/* Copyright 2017 The OpenXLA Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+// SPDX-License-Identifier: Apache-2.0
+
+// Copyright (C) 2024 Intel Corporation
+
+==============================================================================*/
+
+#ifndef XLA_SERVICE_GRAPHCYCLES_GRAPHCYCLES_H_
+#define XLA_SERVICE_GRAPHCYCLES_GRAPHCYCLES_H_
+
+#include <vector>
+
+// GraphCycles detects the introduction of a cycle into a directed
+// graph that is being built up incrementally.
+//
+// Nodes are identified by small integers.  It is not possible to
+// record multiple edges with the same (source, destination) pair;
+// requests to add an edge where one already exists are silently
+// ignored.
+//
+// It is also not possible to introduce a cycle; an attempt to insert
+// an edge that would introduce a cycle fails and returns false.
+//
+// GraphCycles uses no internal locking; calls into it should be
+// serialized externally.
+
+// Performance considerations:
+//   Works well on sparse graphs, poorly on dense graphs.
+//   Extra information is maintained incrementally to detect cycles quickly.
+//   InsertEdge() is very fast when the edge already exists, and reasonably fast
+//   otherwise.
+//   FindPath() is linear in the size of the graph.
+// The current implementation uses O(|V|+|E|) space.
+
+// #include <optional>
+
+// #include "absl/types/span.h"
+// #include <span>
+#include <string>
+
+namespace xla {
+
+// NOTE!!!
+// For now a copy of this is forked to net/plaque. If you
+// find a bug or add a feature, please inform the owners of the
+// net/plaque copy in case it should be integrated.
+// NOTE!!!
+class GraphCycles {
+ public:
+  GraphCycles();
+  ~GraphCycles();
+
+  // Allocate an unused node id and return it.
+  // The new node has a null pointer for its node data.
+  // All node identifiers passed to other routines in this interface
+  // must have been allocated by NewNode() and not yet deallocated
+  // by RemoveNode().
+  int32_t NewNode();
+
+  // Remove "node" from the graph, deleting all edges to and from it.
+  // After this call the identifier "node" it may no longer be used
+  // as an argument to any routine until it has been reallocated with
+  // NewNode().
+  void RemoveNode(int32_t node);
+
+  // Attempt to insert an edge from source_node to dest_node.  If the
+  // edge would introduce a cycle, return false without making any
+  // changes. Otherwise add the edge and return true.
+  bool InsertEdge(int32_t source_node, int32_t dest_node);
+
+  // Remove any edge that exists from source_node to dest_node.
+  void RemoveEdge(int32_t source_node, int32_t dest_node);
+
+  // Return whether there is an edge directly from source_node to dest_node.
+  bool HasEdge(int32_t source_node, int32_t dest_node) const;
+
+  // Contracts the edge from 'a' to node 'b', merging nodes 'a' and 'b'. One of
+  // the nodes is removed from the graph, and edges to/from it are added to
+  // the remaining one, which is returned. If contracting the edge would create
+  // a cycle, does nothing and return no value.
+  int32_t ContractEdge(int32_t a, int32_t b, bool &success);
+
+  // Return true if can contract edge, otherwise return false.
+  bool CanContractEdge(int32_t a, int32_t b);
+
+  // Return whether dest_node is reachable from source_node
+  // by following edges.
+  bool IsReachable(int32_t source_node, int32_t dest_node) const;
+
+  // A faster non-thread-safe version of IsReachable.
+  bool IsReachableNonConst(int32_t source_node, int32_t dest_node);
+
+  // Return or set the node data for a node.  This data is unused
+  // by the implementation.
+  void *GetNodeData(int32_t node) const;
+  void SetNodeData(int32_t node, void *data);
+
+  // Find a path from "source" to "dest".  If such a path exists, place the
+  // node IDs of the nodes on the path in the array path[], and return the
+  // number of nodes on the path.  If the path is longer than max_path_len
+  // nodes, only the first max_path_len nodes are placed in path[].  The client
+  // should compare the return value with max_path_len" to see when this
+  // occurs.  If no path exists, return 0.  Any valid path stored in path[]
+  // will start with "source" and end with "dest".  There is no guarantee that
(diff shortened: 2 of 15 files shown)


source/layers/validation/checkers/events_deadlock/zel_events_deadlock_checker.h
```diff
@@ -0,0 +1,57 @@
...
+      private:
+        // events point from/out to a DAG node. This map stores the DAG ID for each event (if there is one).
+        std::unordered_map<ze_event_handle_t, int> eventToDagID;
+        
+        std::unordered_map<int, actionAndSignalEvent> dagIDToAction;
+
+        // temporary solution to assign unique ID to each DAG. Eventually, this will come from the DAG object that manages the topological sort.
+        int nextDagID = 0;
+    };
+    // class ZESeventsDeadlockChecker : public ZESValidationEntryPoints {};
```
nrspruit, line 51: you still need to define these otherwise you will crash when these are accessed. they just don't need a definition for any of the functions since they are defaulted to empty functions.

bibrak: Done

source/layers/validation/checkers/events_deadlock/zel_events_deadlock_checker.cpp
```diff
@@ -0,0 +1,256 @@
...
+                                             ///< if `nullptr == phWaitEvents`
+    ze_event_handle_t *phWaitEvents          ///< [in][optional][range(0, numWaitEvents)] handle of the events to wait
+                                             ///< on before launching
+) {
+
+    return ZE_RESULT_SUCCESS;
+}
+
+ze_result_t
+eventsDeadlockChecker::ZEeventsDeadlockChecker::zeCommandListAppendMemoryCopyPrologue(
```
bmyates, line 159: I would be insterested in seeing a high level outline of how the algorithm works before getting to implementation of specific functions
1) Which APIs will add things to graph
2) Which APIs will remove things from graph
3) Which APIs will execute deadlock detection mechanism

source/layers/validation/checkers/events_deadlock/zel_events_deadlock_checker.cpp
```diff
@@ -0,0 +1,256 @@
...
+    ze_command_list_handle_t hCommandList, ///< [in] handle of command list
```
bmyates, line 160: I think you may need to consider the specific properties of the command list to determine what to do with it. 

For instance, a command list with `ZE_COMMAND_LIST_FLAG_IN_ORDER` has an implicit dependency between all things appended to it. 

An immediate command list executes immediately and you probably want to do deadlock deetection immediately

a non-immediate command list does not execute until it is explicitly submittted, and you may want to delay deadlock detection until that point. (Or maybe not -not really sure on this one!)

source/layers/validation/checkers/events_deadlock/zel_events_deadlock_checker.cpp
```diff
@@ -0,0 +1,256 @@
...
+        printf("\t\tphWaitEvents[%d] pointer address: %p\n", i, phWaitEvents[i]);
+    }
+
+    int this_action_new_node_id = invalidDagID;
+
+    if (hSignalEvent != nullptr) {
+
+        if (eventToDagID.find(hSignalEvent) != eventToDagID.end()) {
+            if (eventToDagID[hSignalEvent] != invalidDagID) {
+                // This event already exists in the DAG. Get the DAG node ID.
```
bmyates, line 183: just a nitpick, it's not really a DAG if there is a deadlock - because then there is a cycle.

samples/zello_events_deadlock/zello_events_deadlock.cpp
```diff
@@ -0,0 +1,266 @@
...
+    device_desc.stype = ZE_STRUCTURE_TYPE_DEVICE_MEM_ALLOC_DESC;
+    device_desc.pNext = nullptr;
+    device_desc.ordinal = 0;
+    device_desc.flags = 0;
+
+    void *device_mem_ptr = nullptr;
+    SUCCESS_OR_TERMINATE(zeMemAllocDevice(context, &device_desc, buffer_size, 0, pDevice, &device_mem_ptr));
+
+    std::cout << "\n\n\n";
+
```
bmyates, line 200: There is an api `zeEventHostSignal` that allows signalling events from host.   If that API were called right here for event[2], then everything would work.    I think the deadlock detection mechanism needs to account for this API in some way

source/layers/validation/checkers/events_deadlock/zel_events_deadlock_checker.cpp
bibrak: The deadlock detection happens as the API (action) is called from the user app (doesn't matter how the driver stages it in queues). The detection is only based on the `hSignalEvent`, and any `phWaitEvents` events.

source/layers/validation/checkers/events_deadlock/zel_events_deadlock_checker.cpp
```diff
@@ -0,0 +1,256 @@
...
+                    eventToDagID[phWaitEvents[i]] = new_future_node_id;
+
+                    dagID = new_future_node_id;
+                }
+
+                // Add edge (dependency) from dagID to this_action_new_node_id in the DAG.
+                std::cout << "\tDAG: Adding edge from " << dagID << " to " << this_action_new_node_id << std::endl;
+            } else {
+                std::cerr << "eventsDeadlockChecker: zeCommandListAppendMemoryCopyPrologue: Error: Wait event not found in eventToDagID map" << std::endl;
+                std::terminate();
```
bmyates, line 225: If something goes wrong we sohuld return an error instead of terminating.

We also need to make sure any error messages we log are understandable to someone that is not familiar with the internals here

source/layers/validation/checkers/events_deadlock/zel_events_deadlock_checker.cpp
bibrak: Yes, when a cycle is added the underlying DAG structure will not accept that `.add()` and will return error. This way we know there is a cycle and hence a deadlock.

samples/zello_events_deadlock/zello_events_deadlock.cpp
bibrak: But even in that case won't it be an ill formed use of the L0 API? Like a single action is being signaled twice?

source/layers/validation/checkers/events_deadlock/zel_events_deadlock_checker.h
```diff
@@ -0,0 +1,57 @@
...
+        ze_result_t zeCommandListAppendLaunchCooperativeKernelEpilogue(ze_command_list_handle_t hCommandList, ze_kernel_handle_t hKernel, const ze_group_count_t *pLaunchFuncArgs, ze_event_handle_t hSignalEvent, uint32_t numWaitEvents, ze_event_handle_t *phWaitEvents) override;
+        ze_result_t zeCommandListAppendMemoryCopyPrologue(ze_command_list_handle_t hCommandList, void *dstptr, const void *srcptr, size_t size, ze_event_handle_t hSignalEvent, uint32_t numWaitEvents, ze_event_handle_t *phWaitEvents) override;
+        ze_result_t zeCommandListAppendMemoryCopyEpilogue(ze_command_list_handle_t hCommandList, void *dstptr, const void *srcptr, size_t size, ze_event_handle_t hSignalEvent, uint32_t numWaitEvents, ze_event_handle_t *phWaitEvents) override;
+
+      private:
+        // events point from/out to a DAG node. This map stores the DAG ID for each event (if there is one).
+        std::unordered_map<ze_event_handle_t, int> eventToDagID;
```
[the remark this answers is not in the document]
bibrak: Yes, I have yet to recycle them. I will have to come up with something like `eventToDagID[event] = invalidDagID;` when there is a reset called or myriad of those ways that could happen that I am not aware of and will need help later.

rwmcguir (changes requested):
We need the original Apache 2.0 LICENSE file from the XLA project put into the thirdparty/xla folder

third_party/xla/graphcycles.h
```diff
@@ -0,0 +1,167 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+// SPDX-License-Identifier: Apache-2.0
+
+// Copyright (C) 2024 Intel Corporation
```
rwmcguir, line 17: @lisanna-dettwyler a question in our other normal source files we have nomenclature like
//* @file graphcycles.h
in our comments. Should we insert these? 
from legal perspective this is not necessary, but is infrastructure using this for anyreason?

rwmcguir (commented):
From Licensing perspective this looks good.
You have my informal +1 on License basis...

source/layers/validation/checkers/events_deadlock/zel_events_deadlock_checker.cpp
```diff
@@ -0,0 +1,587 @@
...
+ze_result_t
+eventsDeadlockChecker::ZEeventsDeadlockChecker::zeEventDestroyEpilogue(
+    ze_event_handle_t hEvent ///< [in][release] handle of event object to destroy
+) {
+    if (eventToDagID.find(hEvent) != eventToDagID.end()) {
+        // Delete event from eventToDagID but not from the dagIDToAction map as it may be needed for printing the discription of the action when printing path in the DAG.
+        eventToDagID.erase(hEvent);
+    }
+
+    return ZE_RESULT_SUCCESS;
```
aravindksg, line 57: general question: how is epilogue call handled? if the call to driver returns an error then does epilogue get called? also does epilogue need to return the return value from the driver call?

source/layers/validation/checkers/events_deadlock/zel_events_deadlock_checker.h
```diff
@@ -0,0 +1,94 @@
...
+#include "ze_api.h"
+#include "ze_validation_layer.h"
+
+#include <limits>
+#include <string>
+
+namespace validation_layer {
+
+constexpr uint32_t invalidDagID = std::numeric_limits<uint32_t>::max();
+constexpr ze_event_handle_t invalidEventAddress = std::numeric_limits<ze_event_handle_t>::max();
```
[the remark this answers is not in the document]
bibrak: I think the max value of the pointer?

source/layers/validation/README.md
```diff
@@ -48,6 +49,9 @@ This mode maintains an internal mapping of each handle type to a state structure
 - Additional per handle state checks added as needed
     - Example - Check ze_cmdlist_handle_t open or closed
 
+### `ZEL_ENABLE_EVENTSDEADLOCK_CHECKER`
+
+The Events Deadlock Checker is designed to detect potential deadlocks that might occur due to improper event usage in the Level Zero API. It prints out wairning messages for user when it detects a potential deadlock.
```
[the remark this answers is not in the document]
bibrak: Fixed

source/layers/validation/README.md
```diff
@@ -18,6 +18,7 @@ By default, no validation modes will be enabled. The individual validation modes
 
 - `ZE_ENABLE_PARAMETER_VALIDATION`
 - `ZE_ENABLE_HANDLE_LIFETIME`
+- `ZEL_ENABLE_EVENTSDEADLOCK_CHECKER`
```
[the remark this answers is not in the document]
bibrak: That is a good idea.

source/layers/validation/checkers/events_deadlock/zel_events_deadlock_checker.cpp
aravindksg: @nrspruit do you know?

nrspruit: regardless of the result, the epilogue will be called like here:
https://github.com/oneapi-src/level-zero/blob/master/source/layers/validation/ze_valddi.cpp#L49 .

Perhaps we could expand this to pass in the "result" to the epilogue such that a check does not occur or don't call epilogue given an error?

bibrak: Yes, either not call epilogue altogether if error in the actual zeCall or pass the error in the epilogue and then its as per the checker logic and behavior what it needs to do in case of error. 

In the later case, then it will have to change API for all checkers for all zeAPI calls. Or just for the ZeEventDestroy()? 

Again, in the former case it will be only for ZeEventDestroy() or all ZeAPI calls?

source/layers/validation/README.md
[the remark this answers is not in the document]
bibrak: Done!

rwmcguir (approved):

nrspruit:
Depends on https://github.com/oneapi-src/level-zero/pull/235

nrspruit:
@rwmcguir can you sign off if this license is approved?

aravindksg (approved):

rwmcguir (approved):
/ip+1  < haha, this infrastructure doesn't honor this, but just for fun >

nrspruit (approved):

PR #227, opened 2024-10-30
Fix/Regenerate the Loader files incorrectly reordered by the scripts.

nrspruit:
The previous script gen accidently reordered the layers/loader files incorrectly and did not put them back.

This fix correctly addresses the invalid reorder to match the updated script generation.


rwmcguir (approved):

source/layers/tracing/ze_tracing_cb_structs.h
lisanna-dettwyler, line 1: What generates this file?

scripts/input.json
lisanna-dettwyler, line 1: This file shouldn't even be checked in

source/layers/tracing/ze_tracing_cb_structs.h
nrspruit: https://github.com/oneapi-src/level-zero/blob/master/scripts/templates/tracing/trc_cb_struct.h.makov

scripts/input.json
lisanna-dettwyler: It shouldn't be checked in to either repository since it's an artifact, can you just delete it from your branch?

nrspruit: yes, I can just remove it.

nrspruit: Done, removed.

lisanna-dettwyler (approved):

rwmcguir (approved):

aravindksg (approved):

PR #229, opened 2024-10-30
Fix ABI checker build/tests

nrspruit:


rwmcguir (approved):

rwmcguir (approved):

.github/workflows/check-abi.yml
```diff
@@ -133,7 +131,7 @@ jobs:
       run: |
         ./abi-dumper.pl \
           ./head/build/lib/libze_loader.so \
-          -lver $(cat ./head/build/VERSION) \
+          -lver $(cat ./head/build/VERSION)-1 \
```
lisanna-dettwyler, line 134: What effect does adding this have?

.github/workflows/check-abi.yml
```diff
@@ -108,19 +107,18 @@ jobs:
...
         cd base/build
-        cmake -DCMAKE_BUILD_TYPE=Debug ..
+        cmake -E env CXXFLAGS="-Og -g" cmake -DCMAKE_BUILD_TYPE=Debug ..
         make -j$(nproc)
     - name: Build debug head
       run: |
         mkdir head/build
         cd head/build
-        cmake -DCMAKE_BUILD_TYPE=Debug ..
+        cmake -E env CXXFLAGS="-Og -g" cmake -DCMAKE_BUILD_TYPE=Debug ..
```
lisanna-dettwyler, line 116: Why is cmake being passed as an argument to cmake?

.github/workflows/check-abi.yml
nrspruit: Without this the ABI tool had the same version for both versions it was checking which made it think it was checking the same version thus no differences. You need to set some version difference between the two otherwise it does not do a proper ABI check.

.github/workflows/check-abi.yml
nrspruit: This is because a change in the cmake to "add" new flags will not fix the older commit hash that needs to have these flags as well. The ABI Dump tool "requires" that you set these flags otherwise it cannot check the library files and read the symbols. This is a method to pass those flags to the cmake without having to modify the cmake to add the needed flags.

I can confirm that this correctly passes the flags to the cmake build such that both head and base library builds correctly have the needed flags for the ABI check.

lisanna-dettwyler (approved):

PR #230, opened 2024-10-31
Fix metric types for new experimental types

nrspruit:


lisanna-dettwyler (dismissed):

nrspruit:
Abi failure is expected for this patch since it is fixing an issue with ABI in v1.18.4.

lisanna-dettwyler (approved):

PR #231, opened 2024-11-06
Fix backwards compatibility usage of Get*ProcAddrTable

### LINKED ISSUE #196: Driver not initialized: ZE_RESULT_ERROR_UNINITIALIZED
I'm trying to run the zello_world example on Windows 10 and it fails:

```
zello_world.exe
ZE_LOADER_DEBUG_TRACE:Using Loader Library Path:
ZE_LOADER_DEBUG_TRACE:Loading Driver C:\WINDOWS\System32\DriverStore\FileRepository\iigd_dch.inf_amd64_eb1c0bb4c867d89a\ze_intel_gpu64.dll
ZE_LOADER_DEBUG_TRACE:Tracing Layer Library Path: ze_tracing_layer.dll
Driver not initialized: ZE_RESULT_ERROR_UNINITIALIZED
Did NOT find matching ZE_DEVICE_TYPE_GPU device!
```

I have Intel HD Graphics 630 (Kaby-Lake-H-GT2, Gen. 9.5) with the newest intel drivers.
(31.0.101.2130). Its supported according to the list (https://github.com/intel/compute-runtime?tab=readme-ov-file#supported-platforms)

What is exactly missing or in wrong location?

edit:

Debugged and found out that during initialization 
zeDdiTableInit() calls zeGetRTASBuilderExpProcAddrTable() and
it fails because `GET_FUNCTION_PTR( drv.handle, "zeGetRTASBuilderExpProcAddrTable") )` is always null.

Any idea?

lisanna-dettwyler:
Make the results of certain Get*ProcAddrTable calls in ze_libddi.cpp ignored as not all drivers will implement these functions. These functions populate different sections of DDI tables in the lib component.

Fixes #196
Relates to VLCLJ-2339

### DIFF
--- a/source/lib/ze_libddi.cpp
+++ b/source/lib/ze_libddi.cpp
@@ -30,16 +30,18 @@ namespace ze_lib
 
         if( ZE_RESULT_SUCCESS == result )
         {
+            // Optional
             auto getTable = reinterpret_cast<ze_pfnGetRTASBuilderExpProcAddrTable_t>(
                 GET_FUNCTION_PTR(loader, "zeGetRTASBuilderExpProcAddrTable") );
-            result = getTable( ZE_API_VERSION_CURRENT, &initialzeDdiTable.RTASBuilderExp );
+            getTable( ZE_API_VERSION_CURRENT, &initialzeDdiTable.RTASBuilderExp );
         }
 
         if( ZE_RESULT_SUCCESS == result )
         {
+            // Optional
             auto getTable = reinterpret_cast<ze_pfnGetRTASParallelOperationExpProcAddrTable_t>(
                 GET_FUNCTION_PTR(loader, "zeGetRTASParallelOperationExpProcAddrTable") );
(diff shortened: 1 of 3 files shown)


lisanna-dettwyler:
This will support drivers implementing only spec 1.0 functions

nrspruit:
@lisanna-dettwyler ,

Can you do the same in the zesDdiTableInit and zetDdiTableInit because the error will occur there too.

rwmcguir:
Just double checking, this isn't something that needs added to the Mako scripts?

nrspruit (approved):

nrspruit (changes requested):
Add the logic to generate this code from https://github.com/oneapi-src/level-zero/blob/master/scripts/templates/libddi.cpp.mako

nrspruit:
> Just double checking, this isn't something that needs added to the Mako scripts?

Yes, I just realized, this will need to be changed here such that the code gen will create this:
https://github.com/oneapi-src/level-zero/blob/master/scripts/templates/libddi.cpp.mako

nrspruit (approved):

PR #232, opened 2024-11-09
Fix zeInit Compatibility when zeInitDrivers is undefined

nrspruit:
- Handle when zeinitDrivers is called first and the driver does not support the api yet. Included a unit test to cover these types of cases going forward.

### DIFF
--- a/source/lib/ze_libapi.cpp
+++ b/source/lib/ze_libapi.cpp
@@ -197,6 +197,10 @@ zeInitDrivers(
         return result;
     });
 
+    if (result != ZE_RESULT_SUCCESS) {
+        return result;
+    }
+
     if(ze_lib::context->inTeardown) {
         return ZE_RESULT_ERROR_UNINITIALIZED;
     }
--- a/test/loader_api.cpp
+++ b/test/loader_api.cpp
@@ -145,7 +145,7 @@ TEST(
 
 TEST(
     LoaderInit,
-    GivenLevelZeroLoaderPresentWhenCallingZeInitDriversThenzeInitWithzeInitDriversUnsupportedOnTheDriverThenzeInitSucceeds) {
+    GivenZeInitDriversUnsupportedOnTheDriverWhenCallingZeInitDriversThenUninitializedReturned) {
 
   uint32_t pCount = 0;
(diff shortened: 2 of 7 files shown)


test/loader_api.cpp
```diff
@@ -143,6 +143,19 @@ TEST(
   EXPECT_EQ(ZE_RESULT_SUCCESS, zeInit(ZE_INIT_FLAG_GPU_ONLY));
 }
 
+TEST(
+    LoaderInit,
+    GivenLevelZeroLoaderPresentWhenCallingZeInitDriversThenzeInitWithzeInitDriversUnsupportedOnTheDriverThenzeInitSucceeds) {
```
lisanna-dettwyler, line 148: I think a better name might be GivenZeInitDriversUnsupportedOnTheDriverWhenCallingZeInitDriversThenUninitializedReturned since that's the main point of the test.

source/lib/ze_libapi.cpp
```diff
@@ -197,6 +197,10 @@ zeInitDrivers(
         return result;
     });
 
+    if (result != ZE_RESULT_SUCCESS) {
+        return result;
```
lisanna-dettwyler, line 201: I don't see zeInit with this outer return, why does it not need it but here we do?

nrspruit: Let me see if we can check within the init instead, that should result in the same behavior as zeInit.

nrspruit: Either or results in the same behavior where the result of the context init is verified. This was missed here in my case.

nrspruit: nevermind, the reason why this needs to be checked here is because zeInit did not need to continue to allow the L0 apis to work if it failed, zeInitDrivers needs to allow the L0 apis to work so the additional check needs to be made here for the new api until 2.0.

nrspruit: By always checking the result for zeInitDrivers here allows for the driver to be labelled "valid" even if zeInit works and zeInitDrivers does not.

test/loader_api.cpp
nrspruit: I will update the test name.

lisanna-dettwyler (approved):

PR #233, opened 2024-11-12
Fix to add missing zeKernelGetExp API and header updates

nrspruit:


aravindksg (commented):
changes lgtm, so we are manually adding this API due to the spacing issue in the yml file in the spec right? but we need to fix that spacing issue for future releases of spec? and if we fix in yml file then will we have any issues with reordering ddi tables / any other compat issues that we have run into recently?

nrspruit:
> changes lgtm, so we are manually adding this API due to the spacing issue in the yml file in the spec right? but we need to fix that spacing issue for future releases of spec? and if we fix in yml file then will we have any issues with reordering ddi tables / any other compat issues that we have run into recently?

This is not manually added, this is generated after the fix I have to the spec that failed to generate this code. I have this in a draft until the fix I have posted is added to the spec repo.

aravindksg (approved):

lisanna-dettwyler (approved):

PR #235, opened 2024-11-14
Add Passing in the Result to Validation Checkers at Epilogue

nrspruit:


bibrak (approved):
Looks good. I was able to use this change in my feature.

aravindksg (approved):

PR #245, opened 2024-12-02
Update to spec 1.12

lisanna-dettwyler:


Jemale (approved):
lgtm

source/loader/ze_ldrddi.cpp
```diff
@@ -4885,6 +4885,166 @@ namespace loader
...
+        ze_result_t result = ZE_RESULT_SUCCESS;
+
+        // extract driver's function pointer table
+        auto dditable = reinterpret_cast<ze_command_list_object_t*>( hCommandList )->dditable;
+        auto pfnAppendSignalExternalSemaphoreExt = dditable->ze.CommandList.pfnAppendSignalExternalSemaphoreExt;
+        if( nullptr == pfnAppendSignalExternalSemaphoreExt )
+            return ZE_RESULT_ERROR_UNINITIALIZED;
+
+        // convert loader handle to driver handle
+        hCommandList = reinterpret_cast<ze_command_list_object_t*>( hCommandList )->handle;
```
bibrak, line 4977: Is it ok to overwrite the function parameter `hCommandList `?

aravindksg (approved):
lgtm

have we run ABI checkers on this to make sure there are no issues with the Sysman VF interface changes? it's a new interface, but just to make sure

nrspruit:
> lgtm
> 
> have we run ABI checkers on this to make sure there are no issues with the Sysman VF interface changes? it's a new interface, but just to make sure

yep, ABI checker is in the CI and the only "issue" was that the define for current version changed, which is not a problem.

PR #248, opened 2024-12-17
fix: potential memory leaks

bibrak:
Fixes some potential memory leaks that were raised by static code analyzer.

### DIFF
--- a/third_party/xla/ordered_set.h
+++ b/third_party/xla/ordered_set.h
@@ -56,7 +56,10 @@ class OrderedSet {
   // set.
   void Erase(T value) {
     auto it = value_to_index_.find(value);
-    assert(it != value_to_index_.end());
+    if (it == value_to_index_.end()) {
+      std::cerr << "Value not found in OrderedSet" << std::endl;
+      exit(0);
+    }
 
     auto index = it->second;
     // Since we don't want to move values around in `value_sequence_` we swap
--- a/third_party/xla/ordered_set.h
+++ b/third_party/xla/ordered_set.h
@@ -58,10 +58,11 @@ class OrderedSet {
     auto it = value_to_index_.find(value);
     assert(it != value_to_index_.end());
 
+    auto index = it->second;
     // Since we don't want to move values around in `value_sequence_` we swap
     // the value in the last position and with value to be deleted and then
     // pop_back.
-    value_to_index_[value_sequence_.back()] = it->second;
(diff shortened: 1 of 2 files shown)


third_party/xla/ordered_set.h
```diff
@@ -56,12 +56,16 @@ class OrderedSet {
   // set.
   void Erase(T value) {
     auto it = value_to_index_.find(value);
-    assert(it != value_to_index_.end());
```
Jemale, line 59: I guess coverity was complaining about hitting this assert?

Jemale (approved):
Looks fine to me

third_party/xla/ordered_set.h
```diff
@@ -56,12 +56,16 @@ class OrderedSet {
...
+    if (it == value_to_index_.end()) {
+      std::cerr << "Value not found in OrderedSet" << std::endl;
+      exit(0);
+    }
 
+    auto index = it->second;
```
yoon0805, line 64: it should not be null, as the result of iterator should return .end() anyway, so this may not be necessary; Original code seems to be fine

third_party/xla/ordered_set.h
yoon0805: > I guess coverity was complaining about hitting this assert?

So I am also guessing that Bibrak's modification is correct, right? We need to check this during runtime.

bibrak: `assert` is runtime but somehow perhaps the compiler/pre-processor removes it altogether... maybe

third_party/xla/ordered_set.h
bibrak: yes but the original work doesn't work with Coverity...

bibrak: `it` comes from the `map` `value_to_index_` and to assign its value to its own container might be "self-modification", therefore I placed the value in `index` and then assign it.

third_party/xla/ordered_set.h
rwmcguir: *normally assert() is only compiled in during literal debug builds. Coverity uses release builds for analysis
So it is if this line doesn't exist

rwmcguir:
@yoon0805  @Jemale  do we have any other outstanding comments here? We need to get this in and bibrak is out.

rwmcguir (approved):

PR #256, opened 2025-01-08
Fix libddi table query code generation

nrspruit:
- Fixed the libddi table query code generation to correctly handle the check for which ddi tables are required and which are optional to be set by the driver.

### DIFF
--- a/scripts/templates/helper.py
+++ b/scripts/templates/helper.py
@@ -1709,6 +1709,7 @@ Public:
 def get_class_function_objs_exp(specs, cname):
     objects = []
     exp_objects = []
+    optional = True
     for s in specs:
         for obj in s['objects']:
             is_function = obj_traits.is_function(obj)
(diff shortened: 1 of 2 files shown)


scripts/templates/helper.py
```diff
@@ -1718,9 +1719,11 @@ def get_class_function_objs_exp(specs, cname):
                     exp_objects.append(obj)
                 else:
                     objects.append(obj)
+                if obj.get('version',"1.0") == "1.0":
+                    optional = False
```
rwmcguir, line 1723: We set optional to False here for this 's' in Specs, i.e. once per outer loop.
Prior logic didn't flip flop, between states, so after we detect the first optional = True inside the loop should be resetting back to False? I don't claim to undersatnd this loop enough to know if the optional = True statement at the top, now needs to be moved inside the "for s in Specs:" loop to be correct.

nrspruit: So, this function is called to read all the function objects for a single class. 

We want to have the ddi to be "optional" for all "unless" there are 1.0 apis. If there are 1.0 apis, then that ddi table must be queried successfully from the loader.

so, once 1 api is found for that class that is "1.0", then the ddi is no longer optional at that point.

rwmcguir (approved):

PR #259, opened 2025-01-14
Support for Sorting Drivers based on the devices provided

nrspruit:
- Added support for reporting drivers sorted based on the types of devices returned.
- By default, drivers will be sorted such that the ordering will be:
	- Drivers with Discrete GPUs only - Drivers with Discrete and Integrated GPUs - Drivers with Integrated GPUs - Drivers with Mixed Devices Types (ie GPU + NPU) - Drivers with Non GPU Devices Only
- If ZE_ENABLE_PCI_ID_DEVICE_ORDER is set, then the following ordering is provided:
	- Drivers with Integrated GPUs - Drivers with Discrete and Integrated GPUs - Drivers with Discrete GPUs only - Drivers with Mixed Devices Types (ie GPU + NPU) - Drivers with Non GPU Devices Only

- Expanded the zello_world sample to test both init paths for easy prototyping.

### DIFF
--- a/samples/zello_world/zello_world.cpp
+++ b/samples/zello_world/zello_world.cpp
@@ -62,7 +62,7 @@ int main( int argc, char *argv[] )
     {
         tracing_runtime_enabled = true;
     }
-    if( argparse( argc, argv, "-legacy", "--enable_legacy_init" ) )
+    if( argparse( argc, argv, "-legacy_init", "--enable_legacy_init" ) )
     {
         legacy_init = true;
     }
--- a/samples/zello_world/zello_world.cpp
+++ b/samples/zello_world/zello_world.cpp
@@ -40,6 +40,7 @@ void print_loader_versions(){
 int main( int argc, char *argv[] )
 {
     bool tracing_runtime_enabled = false;
+    bool legacy_init = false;
     if( argparse( argc, argv, "-null", "--enable_null_driver" ) )
     {
         putenv_safe( const_cast<char *>( "ZE_ENABLE_NULL_DRIVER=1" ) );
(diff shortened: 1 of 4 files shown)


samples/zello_world/zello_world.cpp
```diff
@@ -61,13 +62,22 @@ int main( int argc, char *argv[] )
     {
         tracing_runtime_enabled = true;
     }
+    if( argparse( argc, argv, "-legacy", "--enable_legacy_init" ) )
```
rwmcguir, line 65: minor suggestion, name -legacy to -legacy_init as many other future options may be legacy in nature.

source/loader/ze_loader_internal.h
```diff
@@ -26,6 +26,18 @@
 #include "spdlog/spdlog.h"
 namespace loader
 {
+    ///////////////////////////////////////////////////////////////////////////////
```
rwmcguir, line 29: Optional, put a comment here that the ordering here is critical and/or used by sorting algorithms, so care should be used when adding.  Also this should NOT be used anyplace else so we can modify this ENUM  (I see this is internal so should be fine, but let's ensure it stays internal) without ABI issues.

rwmcguir (dismissed):
Other than silly comments, I don't have any commentary on the ordering mechanism. Seems simple and easily modifiable via the ENUM, which is ... well pretty cool.

source/loader/ze_loader_internal.h
```diff
@@ -26,6 +26,18 @@
...
+    /// @brief Driver Type Enumerations
+    typedef enum _zel_driver_type_t
+    {
+        ZEL_DRIVER_TYPE_DISCRETE_GPU= 0,          ///< The driver has Discrete GPUs only
+        ZEL_DRIVER_TYPE_GPU = 1,                  ///< The driver has Heterogenous GPU types
```
JablonskiMateusz, line 34: I see that iGPU can be assumed as slower than dGPU  but why assuming that dGPU only  driver is better than dGPU + iGPU ?

source/loader/ze_loader.cpp
```diff
@@ -135,6 +135,47 @@ namespace loader
         return flags_value;
     }
 
+    bool driverSortComparitor(const driver_t &a, const driver_t &b) {
+        bool pciOrderingRequested = getenv_tobool( "ZE_ENABLE_PCI_ID_DEVICE_ORDER" );
```
JablonskiMateusz, line 139: in case of ZE_ENABLE_PCI_ID_DEVICE_ORDER set I believe that all runtimes should also respect the flag and expose lowest pci id first, so we can store pci_id of first device of each runtime and the sort by that

nrspruit: Fixed in latest, the comparator function now reads the driver's pci device order setting such that the env is only read one time and then the value is set for all drivers.

samples/zello_world/zello_world.cpp
nrspruit: fixed

source/loader/ze_loader_internal.h
nrspruit: This remains internal and is not used outside of the loader nor released. I updated the comment accordingly.

source/loader/ze_loader_internal.h
nrspruit: This is an assumption based on an arch review. This is not guaranteed to always be true which is why customers should always ensure the right device is being used in their workloads and not always assume device 0 is the best.

rwmcguir (approved):

PR #261, opened 2025-01-16
Add script to make a copy of headers in include/level_zero

Maetveis:
This is to allow users consuming the headers from a
clone of the repository to be able to refer to the api headers
the same way as it would be if the level-zero headers were installed.
As in after running this script the following will compile
```c++
#include <level_zero/ze_api.h>
```
when `<path-to-repo>/include` is part of the include directory.

Symlinks or alternatively wrapper header files could be added that would
forward to the previous location, but I stuck with copies, since the
files are generated anyway and any solution outside of full copies might
have corner cases. (Symlinks are not always supported on windows,
the build system might be grepping in the contents of the header etc.).

This PR also does not deprecate the current location of the headers.


nrspruit:
This change is rejected, we don't want to be making such a change when a workaround can be added in the user's environment.

PR #262, opened 2025-01-16
Fix to validation layer param checker for extensions

nrspruit:


nrspruit:
error reporting unsupported argument when running the param checker if we don't update the largest stype allowed.

rwmcguir (approved):

PR #270, opened 2025-02-03
Add API logging to validation layer

igchor:


source/layers/validation/ze_validation_layer.cpp
```diff
@@ -15,6 +15,53 @@ namespace validation_layer
 {
     context_t& context = context_t::getInstance();
 
+    void context_t::initLogger() try {
+        std::string loggingLevel = "error";
+        std::string loggingFormat = "ZE ---> %v";
```
igchor, line 20: Should we use the same default format as ze loader here? ([date] [logger_name] [level] message...)

igchor:
My PR changing the expected log format in SYCL has been approved so it's seems that we don't have to add any alternative formats - we can just always use the same formatting as loader is using.

https://github.com/intel/llvm/pull/16897

README.md
```diff
@@ -58,6 +58,19 @@ user's home directory.
...
+The Level Zero Loader will log all API calls whenever logging level is set to `trace` and
+validation layer is enabled. Following variables need to be set to enable API logging:
+
+`ZEL_ENABLE_LOADER_LOGGING=1`
+
+`ZEL_LOADER_LOGGING_LEVEL=trace`
+
+`ZE_ENABLE_VALIDATION_LAYER=1`
+
+By default logs will be written to the log file, as described above. To print the logs
```
rwmcguir, line 71: Though not incorrect in this review, we may need to update the filename above at line 56 as it appears no longer valid according to current behavior. This may be a good time to update.

source/layers/validation/zes_valddi.cpp
```diff
@@ -2411,15 +2528,17 @@ namespace validation_layer
         zes_fan_handle_t hFan                           ///< [in] Handle for the component.
         )
     {
+        context.logger->log_trace("zesFanSetDefaultMode(hFan)");
```
rwmcguir, line 2531: When we log APIs, should we show the hFan value itself? Instead of just the API signature?

source/utils/logging.cpp
```diff
@@ -7,4 +7,87 @@
  */
 
 #include "logging.h"
-namespace loader {} // namespace loader
\ No newline at end of file
+namespace loader {
+
+std::string to_string(const ze_result_t result) {
```
rwmcguir, line 12: Is there a way we can stringify this with MACROs as opposed to keeping another list we have to maintain? Else we have to modify two locations with potential mismatch with new ERROR codes introduced.

README.md
igchor: Are they incorrect? From what I see, we still have LOADER_LOG_FILE_DIRECTORY defined as ".oneapi_logs" and LOADER_LOG_FILE as "ze_loader.log"

source/utils/logging.cpp
igchor: What is the other place where we have a list like this?

I'm not sure if macro would help much but we could probably generate this funtion from a .mako file.

Anyway, I would prefer to do that in a separate PR, I did not introduce this function - I just moved it from ze_loader.cpp

source/layers/validation/zes_valddi.cpp
nrspruit: Lets take this proposal to a future PR.

source/utils/logging.cpp
nrspruit: Lets take this proposal to a future PR.

nrspruit (approved):

nrspruit (approved):

PR #278, opened 2025-02-11
Fix to teardown check to avoid context

nrspruit:


rwmcguir (approved):

rwmcguir:
No questions, does seem odd we had two variables to carry same definition, this should make safety slightly sooner than before.

PR #284, opened 2025-02-21
Add support for 1.7 spec with the static loader

nrspruit:


source/lib/ze_lib.cpp
```diff
@@ -121,11 +91,14 @@ namespace ze_lib
             debug_trace_message(message, to_string(result));
             return result;
         }
+
         bool zeInitDriversSupport = true;
         const std::string loader_name = "loader";
         for (auto &component : versions) {
             if (loader_name == component.component_name) {
                 version = component.spec_version;
+                std::string message = "ze_lib Context Init() Static Loader Found Loader Version v" + std::to_string(component.component_lib_version.major) + "." + std::to_string(component.component_lib_version.minor) + "." + std::to_string(component.component_lib_version.patch);
```
rwmcguir, line 100: nit: Adjusting case to match other statements in function... (less Caps).. adding 'dynamic' as well
"ze_lib Context Init() static loader found dynamic loader version v"

rwmcguir (approved):
My comment is optional, approving, or I can re-approve if you address comment.

source/lib/ze_lib.cpp
nrspruit: dynamic loader found would not make sense here since the dynamic is the loader found.

I am going to leave this for now and merge.

PR #285, opened 2025-02-22
DDI extension support

nrspruit:
- Support in the L0 Loader for https://oneapi-src.github.io/level-zero-spec/level-zero/latest/core/EXT_Driver_DDIHandles.html#ze-extension-driver-ddi-handles 
- To improve the speed for each call of L0 apis into the correct driver, the new support enables for the driver to allocate a header that stores that driver's ddi tables such that the Loader does not need to perform translation of handles to/from ze_object_t.
- Given a driver supports the new extension, the loader will no longer create the ze_object_t and instead use the packed ddi tables in each handle_t.
- The code has been refactored such that the legacy path will continue to work for the driver that has not converted over without interfering with a driver which now uses the "fast" path.
- This new code path also removes the need for handle translation by the user if the drivers all support the new extension.
- This new path is enabled with ZE_ENABLE_LOADER_DRIVER_DDI_PATH=1, by
  default this new path is disabled.


samples/zello_world/zello_world.cpp
```diff
@@ -116,12 +116,9 @@ int main( int argc, char *argv[] )
 
         for( uint32_t driver = 0; driver < driverCount; ++driver )
         {
+            std::cout << "Driver # " << driver << "\n";
             pDriver = drivers[driver];
             pDevice = findDevice( pDriver, type );
-            if( pDevice )
```
rwmcguir, line 121: So clearly, we're going to find all devices now, and not break out at the first one.... However down at line 127 does this logic still hold, if we skip to the end of the driver list via this loop? Maybe instead we insert a temp variable and check pDevice for valid value, increment a counter, and below )(line 128 here, or 127 in new version) replace if (!pDevice) with a if( found_count == 0) ?

scripts/templates/ldrddi.cpp.mako
```diff
@@ -142,167 +144,62 @@ namespace loader
...
+        if (dditable->isValidFlag == 0)
+            return ${X}_RESULT_ERROR_UNINITIALIZED;
+        // Check that api version in the driver is supported by this version of the API
+        if (dditable->version < ${th.get_version(obj)}) {
+            return ${X}_RESULT_ERROR_UNSUPPORTED_VERSION;
         }
+        auto ${th.make_pfn_name(n, tags, obj)} = dditable->${th.get_table_name(n, tags, obj)}->${th.make_pfn_name(n, tags, obj)};
+        if( nullptr == ${th.make_pfn_name(n, tags, obj)} )
+            return ${X}_RESULT_ERROR_UNINITIALIZED;
         %endif
```
rwmcguir, line 174: Do we have to many %endif here

source/loader/ze_loader_api.cpp
```diff
@@ -97,6 +97,7 @@ zelLoaderTranslateHandleInternal(
     }
 
     bool validHandle = false;
+    *handleOut = handleIn;
```
JablonskiMateusz, line 100: don't we need nullptr check before dereferencing the pointer?

scripts/templates/helper.py
```diff
@@ -1702,6 +1702,41 @@ def get_class_function_objs(specs, cname, minVersion = 0, maxVersion = 9999):
...
+            return "ZE_API_VERSION_1_8"
+        if version == "1.9":
+            return "ZE_API_VERSION_1_9"
+        if version == "1.10":
+            return "ZE_API_VERSION_1_10"
+        if version == "1.11":
+            return "ZE_API_VERSION_1_11"
+        if version == "1.12":
+            return "ZE_API_VERSION_1_12"
+        return "ZE_API_VERSION_1_0"
```
JablonskiMateusz, line 1738: I believe default path should break somehow, instead of just silently passing 1.0. If new api version is added and not handled here it shouldn't compile

source/loader/ze_loader_api.cpp
```diff
@@ -96,48 +96,91 @@ zelLoaderTranslateHandleInternal(
         return ZE_RESULT_SUCCESS;
     }
 
+    bool validHandle = false;
```
JablonskiMateusz, line 99: what about handling invalid handle after the entire `switch`?

JablonskiMateusz (commented):
[Add proper cleanup of DDI Driver Tables](https://github.com/oneapi-src/level-zero/pull/285/commits/b1a43bf6f9af446ea42b3e0eb5e5b0c3ac53d27c) - LGTM

source/loader/ze_loader_api.cpp
nrspruit: I can add that at the start of this api.

source/loader/ze_loader_api.cpp
nrspruit: I am removing the variable because it is not needed and I can simplify the check.

scripts/templates/ldrddi.cpp.mako
nrspruit: yes, unfortunately

samples/zello_world/zello_world.cpp
nrspruit: resolved in the latest

scripts/templates/ldrddi_driver_ddi.cpp.mako
```diff
@@ -0,0 +1,96 @@
...
+from templates import helper as th
+%><%
+    n=namespace
+    N=n.upper()
+
+    x=tags['$x']
+    X=x.upper()
+%>/*
+ *
+ * Copyright (C) 2019-2024 Intel Corporation
```
JablonskiMateusz, line 12:
```suggestion
 * Copyright (C) 2025 Intel Corporation
```

rwmcguir (approved):

PR #288, opened 2025-02-26
Fix driver sorting given driver failures and instrumentation

nrspruit:
- To avoid issues when sorting of the drivers can cause an invalid call during instrumentation after driver failures, then the driver sorting has been moved to occur during the call to zeDriverGet or zeinitDrivers.
- In addition, the driver type is not init to UINT32_MAX in the enum such that all failing drivers are forced to the back of the sorted list.

### DIFF
--- a/source/loader/ze_ldrddi.cpp
+++ b/source/loader/ze_ldrddi.cpp
@@ -55,6 +55,10 @@ namespace loader
 
         uint32_t total_driver_handle_count = 0;
 
+        if (!loader::context->coredriverSortingCompleted) {
+            loader::context->coredriverSortingCompleted = loader::context->driverSorting(&loader::context->zeDrivers, nullptr);
+        }
+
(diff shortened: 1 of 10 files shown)


rwmcguir (approved):
Looks ok to me, let's get one more review.

source/loader/ze_ldrddi.cpp
```diff
@@ -55,6 +55,10 @@ namespace loader
 
         uint32_t total_driver_handle_count = 0;
 
+        if (!loader::context->coredriverSortingCompleted) {
```
JablonskiMateusz, line 58: what if `zeDriverGet` is called again within `driverSorting` function?

nrspruit: Fixed in latest, the recursion is now prevented.

rwmcguir (approved):

PR #290, opened 2025-02-27
Allocate context_t always dynamic and enable delayed destroy of context

nrspruit:


include/loader/ze_loader.h
```diff
@@ -70,6 +70,18 @@ zelLoaderTranslateHandle(
 ZE_DLLEXPORT ze_result_t ZE_APICALL
 zelSetDriverTeardown();
 
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Exported function for informing the loader to delay teardown of its context until the call to zelLoaderContextTeardown(). Only applies during static loader.
```
rwmcguir, line 74: Nit:
Only applies if application linked against static loader.
or
Only applies during static loader usage.

rwmcguir (approved):
I don't love the need for this, but the code looks fine.

include/loader/ze_loader.h
nrspruit: Updated in latest

rwmcguir (approved):

PR #295, opened 2025-03-10
SDK Installer For Windows

nrspruit:
- Added Support for building the Level Zero Windows SDK as an MSI installer using WIX
 - Enabled thru CPACK and the define BUILD_INSTALLER=1
 - includes a packaged script level_zero_sdk_setup.ps1 enabling for customers to change the sdk being used after installation
 - installers support side-by-side installation of multiple versions of SDKs
 - LEVEL_ZERO_V1_SDK_PATH is set in the System Environment variables upon installation such that the path to the last installed SDK can be found. The active SDK can be changed using level_zero_sdk_setup.ps1 after installation.


rwmcguir (approved):
Looks good.

PR #297, opened 2025-03-18
Fix sorting drivers missing function pointers and add multi driver ults

nrspruit:


test/loader_api.cpp
```diff
@@ -251,4 +251,70 @@ TEST(
...
+  uint32_t pInitDriversCount = 0;
+  uint32_t pDriverGetCount = 0;
+  ze_init_driver_type_desc_t desc = {ZE_STRUCTURE_TYPE_INIT_DRIVER_TYPE_DESC};
+  desc.flags = UINT32_MAX;
+  desc.pNext = nullptr;
+  EXPECT_EQ(ZE_RESULT_SUCCESS, zeInitDrivers(&pInitDriversCount, nullptr, &desc));
+  EXPECT_GT(pInitDriversCount, 0);
+  EXPECT_EQ(ZE_RESULT_SUCCESS, zeInit(0));
+  EXPECT_EQ(ZE_RESULT_SUCCESS, zeDriverGet(&pDriverGetCount, nullptr));
+  EXPECT_GT(pDriverGetCount, 0);
```
bibrak, line 301: as we run it with env var `ZE_ENABLE_ALT_DRIVERS=libze_null_test1.so,libze_null_test2.so`, will this `pDriverGetCount == 2` ?

nrspruit: We want to keep the test itself generic, the usage of the alt drivers is in the cmake.

rwmcguir (approved):

PR #308, opened 2025-03-27
Address zesInit issues with error handling and driver sorting.

nrspruit:
- Fix issues where zesInit failing can cause other driver inits to fail. Sysman Only init is now completely isolated.
- Fixed driver sorting when using zesInit to use matched core driver.


bibrak (approved):
LGTM

PR #309, opened 2025-03-31
Restrict Intel VPU driver to ze_intel_npu for Android

yongzhi1:
When both ze_intel_vpu.so and ze_intel_npu.so are present in the file sys at loading time, ze_intel_npu.so will raise an exception and return ZE_RESULT_ERROR_UNINITIALIZED, so let's limit it to single NPU Driver.

### DIFF
--- a/source/loader/linux/driver_discovery_lin.cpp
+++ b/source/loader/linux/driver_discovery_lin.cpp
@@ -18,7 +18,9 @@ namespace loader {
 static const char *knownDriverNames[] = {
     MAKE_LIBRARY_NAME("ze_intel_gpu", "1"),
     MAKE_LIBRARY_NAME("ze_intel_gpu_legacy1", "1"),
+#if !defined(ANDROID)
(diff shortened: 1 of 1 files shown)


nrspruit:
@yongzhi1 , this seems to be the Loader addressing a coexistence bug that the NPU/VPU drivers don't handle correctly.

I am resistant to make driver specific changes like this unless there is no possibility of fixing this issue in the drivers, is that the case?

for instance, the ze_intel_gpu_legacy1 and ze_intel_gpu are able to coexist so I am concerned why we would have to make this exception.

yongzhi1:
> @yongzhi1 , this seems to be the Loader addressing a coexistence bug that the NPU/VPU drivers don't handle correctly.
> 
> I am resistant to make driver specific changes like this unless there is no possibility of fixing this issue in the drivers, is that the case?
> 
> for instance, the ze_intel_gpu_legacy1 and ze_intel_gpu are able to coexist so I am concerned why we would have to make this exception.

Ack, will see if the issue can be handled better outside of level-zero, thus close this PR.

PR #311, opened 2025-04-02
Added zelCheckIsLoaderInTearDown function to check if the loader is i…

nrspruit:
…n teardown

- To enable verification of the loader teardown process, a new function `zelCheckIsLoaderInTearDown` has been added to the `ze_loader.h` header file.
- This function can be used to check if the loader is in the process of tearing down. It returns a boolean value indicating whether the loader is currently in teardown mode or not.

### DIFF
--- a/source/lib/ze_lib.cpp
+++ b/source/lib/ze_lib.cpp
@@ -148,6 +148,7 @@ namespace ze_lib
 
         std::string version_message = "Loader API Version to be requested is v" + std::to_string(ZE_MAJOR_VERSION(version)) + "." + std::to_string(ZE_MINOR_VERSION(version));
         debug_trace_message(version_message, "");
+        loaderDriverGet = reinterpret_cast<ze_pfnDriverGet_t>(GET_FUNCTION_PTR(loader, "zeDriverGet"));
 #else
         result = zeLoaderInit();
         if( ZE_RESULT_SUCCESS == result ) {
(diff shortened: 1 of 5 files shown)


source/lib/ze_lib.cpp
```diff
@@ -401,6 +404,81 @@ zelSetDelayLoaderContextTeardown()
...
 
+#ifdef DYNAMIC_LOAD_LOADER
+void stabilityCheck(std::promise<int> stabilityPromise) {
+    try {
+        if (!ze_lib::context->loader) {
+            if (ze_lib::context->debugTraceEnabled) {
+                std::string message = "Loader module is null. Exiting stability checker thread.";
+                ze_lib::context->debug_trace_message(message, "");
+            }
+            stabilityPromise.set_value(1);
```
rwmcguir, line 415: Can we change the 1, 2, 3 values being set to #defines or enums to give descriptive name?

source/lib/ze_lib.cpp
```diff
@@ -401,6 +404,81 @@ zelSetDelayLoaderContextTeardown()
     #endif
 }
 
+#ifdef DYNAMIC_LOAD_LOADER
+void stabilityCheck(std::promise<int> stabilityPromise) {
```
rwmcguir, line 408: Please add some basic docs around the expectations of the thread and when it would be killed or dies.

nrspruit: updated in the latest

source/lib/ze_lib.cpp
nrspruit: added defines for the return codes.

rwmcguir (approved):

PR #312, opened 2025-04-04
Always check and update the GUID when version changes

nrspruit:


rwmcguir (approved):
Looks smart and actually makes it simpler.

PR #321, opened 2025-04-17
Add fast path for teardown check with a shared thread

nrspruit:


source/lib/ze_lib.cpp
```diff
@@ -1,3 +1,4 @@
+
```
rwmcguir, line 1: White space promotes readership?

source/lib/ze_lib.cpp
```diff
@@ -490,18 +521,60 @@ zelCheckIsLoaderInTearDown() {
...
-        result = resultFuture.get(); // Blocks until the result is available
-        if (ze_lib::context->debugTraceEnabled) {
-            std::string message = "Stability checker thread completed with result: " + std::to_string(result);
-            ze_lib::context->debug_trace_message(message, "");
+        // Launch the stability checker thread on the first call
+        static std::once_flag stabilityThreadFlag;
+        std::lock_guard<std::mutex> lock(*ze_lib::stabilityMutex);
+        *ze_lib::stabilityPromiseResult = std::promise<int>();
+        *ze_lib::resultFutureResult = ze_lib::stabilityPromiseResult->get_future();
+        ze_lib::stabilityCheckThreadStarted->store(1);
```
rwmcguir, line 535: Slight comment for readability, you are using 1 here (likely to indicate RUNNING) and back in the destruction you use -1 as a message to exit the thread. Maybe use define or enums so we can read it easier. :-)
1 = RUNNING
0 = STOPPED? QUISECED? 
-1 = EXIT_THREAD?

rwmcguir (commented):
Overall looks pretty good, my commentary can be approved tomorrow if we really don't want to change anything, since it is not functional.

PR #322, opened 2025-04-22
[DO NOT MERGE] Update to spec v1.13.1

vishnu-khanth:


source/layers/tracing/ze_tracing_cb_structs.h
```diff
@@ -30,6 +30,11 @@ typedef struct _zel_global_callbacks_t
 /// @brief Table of RTASBuilder callback functions pointers
 typedef struct _zel_rtas_builder_callbacks_t
 {
+    ze_pfnRTASBuilderCreateExtCb_t                                  pfnCreateExtCb;
```
rwmcguir, line 33: @nrspruit @vishnu-khanth  I feel this is odd that we're adding members to the start of a structure, shouldn't they be at the end.. Is this weird artifact of this temp branch or did we mess something up with the RTASBuilder extension promotion...... Do we even care with this specific structure if totally internal.

nrspruit: because these are assigned individually, the ordering of the callbacks is not important.

include/ze_api.h
```diff
@@ -1187,7 +1282,7 @@ typedef enum _ze_api_version_t
     ZE_API_VERSION_1_11 = ZE_MAKE_VERSION( 1, 11 ),                         ///< version 1.11
     ZE_API_VERSION_1_12 = ZE_MAKE_VERSION( 1, 12 ),                         ///< version 1.12
     ZE_API_VERSION_CURRENT = ZE_MAKE_VERSION( 1, 12 ),                      ///< latest known version
-    ZE_API_VERSION_FORCE_UINT32 = 0x7fffffff
```
nrspruit, line 1190: this is a bug in the spec scripts, you need to manually add the 1.13 version.

nrspruit: also, you need to update:
https://github.com/oneapi-src/level-zero-spec/blob/master/scripts/core/driver.yml#L213

rwmcguir:
We should close this review... We will repost another 1.13.1 once we get the N-1 N+1 compatiblity check done Neil is working on.

PR #328, opened 2025-04-25
fix ddi compatability to avoid assigning values which don't exist

nrspruit:


source/layers/tracing/ze_trcddi.cpp
```diff
@@ -7751,18 +7751,19 @@ zeGetGlobalProcAddrTable(
...
 
     ze_result_t result = ZE_RESULT_SUCCESS;
 
-    dditable.pfnInit                                     = pDdiTable->pfnInit;
-    pDdiTable->pfnInit                                   = tracing_layer::zeInit;
-
-    dditable.pfnInitDrivers                              = pDdiTable->pfnInitDrivers;
-    pDdiTable->pfnInitDrivers                            = tracing_layer::zeInitDrivers;
-
+    if (version >= ZE_API_VERSION_1_0) {
```
rwmcguir, line 7759: Not wrong, but seems unnecessary to have this if() conditional, everything is newer than 1.0

scripts/templates/ldrddi.cpp.mako
```diff
@@ -443,15 +443,17 @@ ${tbl['export']['name']}(
         {
             // return pointers to loader's DDIs
             %for obj in tbl['functions']:
+            if (version >= ${th.get_version(obj)}) {
```
rwmcguir, line 446: Since this creates one if() per API, it works fine...
We should find a way in later edition to have on if() per version and place API's in proper if
functionally, it won't make a difference, so it can go in today as is, but we can clean this up later.

rwmcguir: That may require nested loops, or make assumptions that the spec itself always passes APIs in order of version... Either way seems like a more complicated design that will take more time that we have at the moment.

rwmcguir (approved):

PR #340, opened 2025-05-28
Init certification checker

marekkozl:


nrspruit (commented):
This looks good, can you add a README.md in the "certification" folder describing the checker and what it does and how to enable it?

also, since this is another validation mode, please add a section about this mode in https://github.com/oneapi-src/level-zero/blob/master/source/layers/validation/README.md#validation-modes

PR #360, opened 2025-07-17
Add tests for Loader Driver interaction conformance for Sysman VF Management API's

vishnu-khanth:
Related-To: NEO-15334

### DIFF
--- a/scripts/templates/nullddi.cpp.mako
+++ b/scripts/templates/nullddi.cpp.mako
@@ -92,6 +92,12 @@ namespace driver
 
             %endfor
         }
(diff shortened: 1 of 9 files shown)


vishnu-khanth:
Can those two failed checks re-run ? Looks like i don't have permission to re-run.

nrspruit (approved):
Looks good to me!

nrspruit (approved):

vishnu-khanth:
Right now keeping it in draft bcos exploring if any other better method exists which can be extensible for other validation like structs in/out.
E.g: Instead of env, writing to a file etc...

Will finalize before merging this & going for updating all API's

PR #361, opened 2025-07-21
Enable DDI Driver Path by Default

nrspruit:


source/loader/ze_loader.cpp
```diff
@@ -459,7 +459,9 @@ namespace loader
         loader::loaderZetDdiTable = new zet_dditable_t();
         loader::loaderZesDdiTable = new zes_dditable_t();
         debugTraceEnabled = getenv_tobool( "ZE_ENABLE_LOADER_DEBUG_TRACE" );
-        driverDDIPathDefault = getenv_tobool( "ZE_ENABLE_LOADER_DRIVER_DDI_PATH" );
+        // Set to true if the env is not set
+        std::string ddiPathConfig = getenv_string("ZE_ENABLE_LOADER_DRIVER_DDI_PATH");
+        driverDDIPathDefault = (ddiPathConfig.empty()) ? true : getenv_tobool("ZE_ENABLE_LOADER_DRIVER_DDI_PATH");
```
rwmcguir: doesn't getenv_tobool read the environment again for the same variable? 
Should this be something like 
driverDDIPathDefault = (ddiPathConfig.empty()) ? true : ddiPathConfig.<somefunc_that returns true/false without re-reading environ)

at scale these calls are expensive in load time

nrspruit: So, you want a string to bool function call?

nrspruit: updated in latest

rwmcguir (approved):

source/loader/ze_loader.cpp
rwmcguir: I didn't have a specific target just don't read environ twice for the same thing.

PR #363, opened 2025-07-23
fix: Avoid invalid casting into loader objects when DDI ext is supported

vishnu-khanth:
Related-To: NEO-15615

### DIFF
--- a/source/loader/ze_loader_api.cpp
+++ b/source/loader/ze_loader_api.cpp
@@ -103,11 +103,11 @@ zelLoaderTranslateHandleInternal(
     switch(handleType){
         case ZEL_HANDLE_DRIVER:
         {
-            // Determine if legacy loader interception is enabled or 
-	    // loader object exists for that handle by checking two conditions:
-            // 1. DDI Ext is disabled in loader OR
-            // 2. The input handle's function pointer table matches the loader's dispatch table,
-            //    indicating this handle was created through the loader's intercept layer
+            // Determine if legacy loader intercept is enabled 
+	    // i.e., loader object exists for that handle by checking two conditions:
+            // 1. DDI Ext is disabled in loader by default OR
+            // 2. The input handle's function pointer matches the loader's dispatch table pointer,
+            //    indicating that the handle was created through the legacy loader intercept layer
 	    
 	    bool legacy_ldr_intercept_enabled = (!loader::context->driverDDIPathDefault) ||
                                                 (reinterpret_cast<ze_handle_t *>(handleIn)->pCore->Driver->pfnGet ==


source/loader/ze_loader_api.cpp
```diff
@@ -101,84 +101,170 @@ zelLoaderTranslateHandleInternal(
...
     *handleOut = handleIn;
     switch(handleType){
-        case ZEL_HANDLE_DRIVER: 
-            if (loader::context->ze_driver_factory.hasInstance(reinterpret_cast<loader::ze_driver_object_t*>(handleIn)->handle)) {
-                *handleOut = reinterpret_cast<loader::ze_driver_object_t*>( handleIn )->handle;
+        case ZEL_HANDLE_DRIVER:
+        {
+            bool legacy_ldr_intercept_enabled = (!loader::context->driverDDIPathDefault) ||
+                                                (reinterpret_cast<ze_handle_t *>(handleIn)->pCore->Driver->pfnGet ==
+                                                 loader::loaderDispatch->pCore->Driver->pfnGet);
```
nrspruit, line 114: Can you add a comment here that what you are doing is:

1) checking if by default we have the DDI Ext disabled and 2) checking if the handleIn matches the backwards compatible loader::loaderDispatch tables?

that way when someone looks at this they can understand what is being checked?

also, do you have an example where this failed recently?

vishnu-khanth: Hi @nrspruit, will update the jira NEO-15615 with detailed analysis & reproducer for this issue.

joshuaranjan (commented):
@vishnu-khanth 
Could we add tests here to validate this issue does not exist and works OK ?

source/loader/ze_loader_api.cpp
vishnu-khanth: Updated with comments

vishnu-khanth: @nrspruit 
How about having only one check in beginning before switch and return directly?
```
(!loader::context->driverDDIPathDefault) || (reinterpret_cast<ze_handle_t *>(handleIn)->pCore->Driver->pfnGet == loader::loaderDispatch->pCore->Driver->pfnGet)
```
                                                 
                                                 
 Instead of having for all different types of handles?

nrspruit: Because we are checking based on each handle type. You would still have to check which handle type to avoid casting a handle to a different type. This is cleaner and I will merge this change since it resolves the issue. We can work on possibly improving this later, but once all drivers have migrated to DDI ext, this function is defunct.

nrspruit (approved):

source/loader/ze_loader_api.cpp
vishnu-khanth: Ok. Thank you!

vishnu-khanth:
Looks like it got added as multi commits in oneapi-src:master

PR #371, opened 2025-09-15
Support for ZEL_DRIVERS_ORDER to order based off user input

nrspruit:


test/driver_ordering_helper_tests.cpp
```diff
@@ -0,0 +1,408 @@
...
+        return loader::ZEL_DRIVER_TYPE_GPU;
+    } else if (typeStr == "INTEGRATED_GPU_ONLY") {
+        return loader::ZEL_DRIVER_TYPE_INTEGRATED_GPU;
+    } else if (typeStr == "NPU") {
+        return loader::ZEL_DRIVER_TYPE_NPU;
+    }
+    return loader::ZEL_DRIVER_TYPE_FORCE_UINT32; // Invalid
+}
+
+// Helper function to trim whitespace (copy from implementation)
```
rwmcguir: Very Minor: In the future, would it would make more sense to have a util.h or something with these things defined, since we're duplicating functions between code bodies.

nrspruit: fixed in latest

rwmcguir (approved):

PR #377, opened 2025-10-13
Update Loader scripts and source to support v1.14.33 of the L0 Spec

nrspruit:
* fix: Update scripts to handle dynamic return type for source/lib
* Update null driver scripts to have dynamic return type and zer API's
* Update scripts to handle dynamic return types for source/loader & default DDI table for zer API's
* Update tracing layer scripts to handle dynamic return type of API's
* Update Validation layer to handle dynamic return types of API's
* Integrate zer namespace in null driver
* Integrate zer namespace in source/lib
* Integrate zer namespace in source/loader
* Integrate zer namespace in validation layer
* Integrate zer namespace in tracing layer
* Track the DriverHandle per Driver as zerDriverHandle, set to nullptr given the driver supports DDI Driver EXT, but does not support ZER APIs.
* Enable support for DDI Driver Ext flag etc checks given a handle.
* Backwards compatability guranteed with ze_object.h creation with pRuntime set in the loader handle.
* Fix default assignment of zerDriver handle and ddi table and refactor
* Add thread local storage error description for zer API's


include/ze_api.h
```diff
@@ -2965,7 +3045,7 @@ zeCommandListCreate(
 ///         + `nullptr == altdesc`
 ///         + `nullptr == phCommandList`
 ///     - ::ZE_RESULT_ERROR_INVALID_ENUMERATION
-///         + `0x3 < altdesc->flags`
+///         + `0x7 < altdesc->flags`
```
rwmcguir, line 3048: Why would this value be changing?... It appears to just be comment, but I am not seeing why this value would shift.

rwmcguir: I figured it out... just a value in total number of flag types...

source/loader/ze_ldrddi.cpp
```diff
@@ -279,7 +339,7 @@ namespace loader
         )
     {
         ze_result_t result = ZE_RESULT_SUCCESS;
-
+
```
vishnu-khanth, line 342: minor : Looks like additional indentation/new line difference got added in ze*_ldrddi.cpp files (which existed in inner-source PR as well but missed to correct it)

PR #378, opened 2025-10-16
fix: segfaults in zer DDI intercepts

vishnu-khanth:


source/drivers/null/ze_null.cpp
```diff
@@ -487,17 +487,32 @@ namespace driver
...
+                ze_driver_extension_properties_t ddiHandlesExtension = {};
+#if defined(_WIN32)
+                strcpy_s( ddiHandlesExtension.name, ZE_DRIVER_DDI_HANDLES_EXT_NAME );
+#else
+                strcpy( ddiHandlesExtension.name, ZE_DRIVER_DDI_HANDLES_EXT_NAME );
 #endif
-                driverExtensionProperties.version = ZET_API_TRACING_EXP_VERSION_1_0;
 
-                *pExtensionProperties = driverExtensionProperties;
+                auto ddi_version_env = getenv("ZEL_TEST_DDI_HANDLES_EXT_VERSION");
```
nrspruit: We have a helper function to get env, you cannot call this directly.

vishnu-khanth: Updated

PR #379, opened 2025-10-16
Refactor Init to delay Library Load

nrspruit:
- Delay to load of the L0 Driver Libraries until the user has directly
  requested the library to be loaded thru the flags.
- Remove the "pretest" of driver support and instead assume driver
  support based on naming and registry categorization. Speeds up loader
  init.
- Moved init of driver libraries until zInitDrivers or zeInit
- Added ULTs to verify the new init_driver functionality

### DIFF
--- a/.github/workflows/build-quick-static-n-1.yml
+++ b/.github/workflows/build-quick-static-n-1.yml
@@ -54,6 +54,7 @@ jobs:
             -D CMAKE_CXX_COMPILER_LAUNCHER=ccache \
             -D CMAKE_BUILD_TYPE=Release \
             -D BUILD_STATIC=0 \
+            -D BUILD_L0_LOADER_TESTS=1 \
             ..
           make -j$(nproc)
 


.github/workflows/build-quick-static.yml
```diff
@@ -35,6 +35,7 @@ jobs:
               -D CMAKE_C_COMPILER_LAUNCHER=ccache \
               -D CMAKE_CXX_COMPILER_LAUNCHER=ccache \
               -D CMAKE_BUILD_TYPE=Release \
+              -D BUILD_L0_LOADER_TESTS=1 \
```
vishnu-khanth, line 38: I think may not be required since BUILD_L0_LOADER_TESTS is enabled for static loader build

test/loader_api.cpp
```diff
@@ -220,7 +220,7 @@ TEST(
   desc.flags = UINT32_MAX;
   desc.pNext = nullptr;
   putenv_safe( const_cast<char *>( "ZEL_TEST_MISSING_API=zeInitDrivers" ) );
-  EXPECT_EQ(ZE_RESULT_ERROR_UNINITIALIZED, zeInitDrivers(&pInitDriversCount, nullptr, &desc));
+  EXPECT_EQ(ZE_RESULT_ERROR_UNSUPPORTED_FEATURE, zeInitDrivers(&pInitDriversCount, nullptr, &desc));
```
vishnu-khanth, line 223: minor: test name could also be updated

test/loader_tracing_layer.cpp
```diff
@@ -978,8 +978,9 @@ namespace
         const char *errorString = nullptr;
         ze_result_t result = zerGetLastErrorDescription(&errorString);
         EXPECT_EQ(ZE_RESULT_ERROR_UNSUPPORTED_FEATURE, result);
-        EXPECT_EQ(0, tracingData.getZerPrologueCallCount("zerGetLastErrorDescription"));
-        EXPECT_EQ(0, tracingData.getZerEpilogueCallCount("zerGetLastErrorDescription"));
+        // ZER callbacks should still be called in the tracing layer even if the driver ends up not supporting ZER APIs
```
vishnu-khanth, line 981: minor: test name can be updated here aswell

.github/workflows/build-quick-static.yml
nrspruit: Because of the fake drivers I have created this indicates the creation of those libraries as part of that build option. so, even though we don't use the "tests" in the dynamic build folder, we will use the test libraries in the lib.

test/loader_api.cpp
nrspruit: Good point, I will update that

test/loader_tracing_layer.cpp
nrspruit: Good point, I will rewrite the test name.

nrspruit: fixed in latest

test/loader_api.cpp
nrspruit: fixed in latest

.github/workflows/build-quick-static.yml
vishnu-khanth: Got it. Then I guess the same thing can be added to buil-quick-static-n-1 aswell.
So that for static n-1 testing for future PR's  (which will have driver init tests as part of n-1 static build), fake libs gets generated in dynamic_loader folder to use it for driver init tests

vishnu-khanth: Other than this, looks good to me!

nrspruit: fixed in latest, thanks!

vishnu-khanth (approved):

rwmcguir (approved):
Looks fine to me, other than the failing content below...

PR #385, opened 2025-10-23
Fix to zer Handle init to address replacement of ze driver handles

nrspruit:
- Removed generation of zerDriver Handles and reuse the ze driver handles when running in legacy ddi table modes.
- Fixed ULT for driverget to properly read handles.


rwmcguir (approved):
Approval but we still need more approval to be sure.

vishnu-khanth (approved):
LGTM!

PR #387, opened 2025-10-28
Add Dynamic Tracing Support for Zer API's

vishnu-khanth:


source/loader/ze_loader_api.h
```diff
@@ -37,10 +37,16 @@ zelLoaderDriverCheck(ze_init_flags_t flags, ze_init_driver_type_desc_t* desc, ze
 
 
 ///////////////////////////////////////////////////////////////////////////////
-/// @brief Internal function for Setting the ddi table for the Tracing Layer.
+/// @brief Internal function for Setting the ZE ddi table for the Tracing Layer.
 ///
 ZE_DLLEXPORT ze_result_t ZE_APICALL
-zelLoaderTracingLayerInit(std::atomic<ze_dditable_t *> &zeDdiTable);
+zelLoaderZeTracingLayerInit(std::atomic<ze_dditable_t *> &zeDdiTable);
```
nrspruit: you cannot rename an already exported symbol, you can only add the new zer symbol

vishnu-khanth: Updated. Will add test to confirm zer dynamic tracing is considered as optional in case where zer API's are not supported and update this PR accordingly.

nrspruit: thank you! the code looks good now.

nrspruit (approved):

PR #396, opened 2025-12-02
feature: Add zelGetLoaderVersion, allowing for query of Loader Version without init

nrspruit:
- New API zelGetLoaderVersion, allows for loader version check without init for dynamic loader usecases.
- zelGetLoaderVersion called from static loader applications will require zeInitDrivers due to requring the loader init to be completed.


rwmcguir (approved):
This looks solid to me and thank you for adding documentation to prior APIs

PR #412, opened 2026-01-14
Add -fno-gnu-unique to compile options for gcc

mateusztabaka:
Unique symbols prevent the library from being dlclosed.


nrspruit (changes requested):
Setting this causes issues with improvements of cross‑DSO C++ behavior and RTLD flags of libraries using L0.

We cannot move ahead with such a change without considerable customer and usecase review, otherwise we risk impacting customers negatively. 

Until we can ensure this will not impact our range of customer usecases outside the dlclose needs, I cannot approve of this change.

rwmcguir:
We are working on an alternative to this change, which will remove unique symbols.... I'll check back when this is ready.

PR #414, opened 2026-01-14
Update Tracing Layer Documentation for the expanded support

nrspruit:
- Fix the Tracing Layer Readme to outline the full feature set as of additions including dynamic tracing enable/disable and query of tracing state and additional clarifications.


source/layers/tracing/README.md
```diff
@@ -164,7 +208,7 @@ void TracingExample2( ... )
     tracer_desc.stype = ZEL_STRUCTURE_TYPE_TRACER_DESC;
     tracer_desc.pUserData = &tracer_data;
     zel_tracer_handle_t hTracer;
-    zelTracerCreate(hDevice, &tracer_desc, &hTracer);
+    zelTracerCreate(&tracer_desc, &hTracer);
```
rwmcguir, line 211: Just making a mental note that this is OK, i.e. removing the hDevice handle as it was not supposed to be here to start with.

PR #415, opened 2026-01-15
Fix closing driver library in context destructor

mateusztabaka:
Loop over zeDrivers (instead of allDrivers) vector in order to free driver libraries. Using allDrivers may not always work since if there's more than one driver discovered, drivers in allDrivers vector have null handles.


nrspruit:
Please see the following PR https://github.com/oneapi-src/level-zero/pull/416 this will address this issue without breaking other scenarios.

PR #417, opened 2026-01-20
fix memory leaks in loader

vishnu-khanth:
Related-To: NEO-17220


vishnu-khanth:
More details with logs attached in NEO-17220

nrspruit (approved):

PR #418, opened 2026-01-22
Loader Driver Interaction conformance tests for Sysman Exp and Ext APIs

pratikbariintel:
Added tests for PciLinkSpeedDowngrade, PowerLimits, RasState, Firmwaresecurityversion, SysmanDeviceMapping functions which includes the specs experimental and the extension APIs.

Related-To: NEO-15551

Signed-off-by: Pratik Bari <pratik.bari@intel.com>


pratikbariintel:
> Can we confirm whether we have 100% function coverage in source/lib/zes_libapi.cpp ?
> 
> If something is not covered, then the test could be added as part of this PR itself ?

I have checked with lib/zes_libapi.cpp and we have 100% Sysman functional coverage

PR #432, opened 2026-02-25
Fix the clang issue due to empty structure in the Tracing Layer

pratikbariintel:
Use void* instead of empty zer_get_default_context_params_t structure for parameter-less callbacks to fix Clang compilation.

Related-To: NEO-17790

Signed-off-by: Pratik Bari <pratik.bari@intel.com>

### DIFF
--- a/include/layers/zel_tracing_register_cb.h
+++ b/include/layers/zel_tracing_register_cb.h
@@ -2721,26 +2721,16 @@ typedef void (ZE_APICALL *zer_pfnTranslateIdentifierToDeviceHandleCb_t)(
     void** ppTracerInstanceUserData
     );
 
-///////////////////////////////////////////////////////////////////////////////
-/// @brief Callback function parameters for zerGetDefaultContext
-/// @details Each entry is a pointer to the parameter passed to the function;
-///     allowing the callback the ability to modify the parameter's value
-
-typedef struct _zer_get_default_context_params_t
-{
-    void* dummy;  // Placeholder for empty parameter list
-} zer_get_default_context_params_t;
-
 
 ///////////////////////////////////////////////////////////////////////////////
 /// @brief Callback function-pointer for zerGetDefaultContext
-/// @param[in] params Parameters passed to this instance
+/// @param[in] params Parameters passed to this instance (NULL for parameter-less functions)
 /// @param[in] result Return value
 /// @param[in] pTracerUserData Per-Tracer user data
 /// @param[in,out] ppTracerInstanceUserData Per-Tracer, Per-Instance user data
 
 typedef void (ZE_APICALL *zer_pfnGetDefaultContextCb_t)(
-    zer_get_default_context_params_t* params,
+    void* params,
     ze_context_handle_t result,
     void* pTracerUserData,
     void** ppTracerInstanceUserData
--- a/include/layers/zel_tracing_register_cb.h
+++ b/include/layers/zel_tracing_register_cb.h
@@ -2728,6 +2728,7 @@ typedef void (ZE_APICALL *zer_pfnTranslateIdentifierToDeviceHandleCb_t)(
 
 typedef struct _zer_get_default_context_params_t
 {
+    void* dummy;  // Placeholder for empty parameter list
 } zer_get_default_context_params_t;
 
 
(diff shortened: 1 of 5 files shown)


scripts/templates/tracing/trc_setters.h.mako
```diff
@@ -44,28 +44,40 @@ typedef struct _zel_tracer_handle_t *zel_tracer_handle_t;
...
+
 /// @param[in] result Return value
 /// @param[in] pTracerUserData Per-Tracer user data
 /// @param[in,out] ppTracerInstanceUserData Per-Tracer, Per-Instance user data
 
 typedef void (${X}_APICALL *${th.make_pfncb_type(n, tags, obj)})(
+    %if param_lines:
     ${th.make_pfncb_param_type(n, tags, obj)}* params,
+    %else:
+    void* params,
```
joshuaranjan, line 79: Can we not have params at all ?
Is void * necessary ?

include/layers/zel_tracing_register_cb.h
```diff
@@ -2721,25 +2721,16 @@ typedef void (ZE_APICALL *zer_pfnTranslateIdentifierToDeviceHandleCb_t)(
...
 /// @brief Callback function-pointer for zerGetDefaultContext
-/// @param[in] params Parameters passed to this instance
+/// @param[in] params Parameters passed to this instance (NULL for parameter-less functions)
 /// @param[in] result Return value
 /// @param[in] pTracerUserData Per-Tracer user data
 /// @param[in,out] ppTracerInstanceUserData Per-Tracer, Per-Instance user data
 
 typedef void (ZE_APICALL *zer_pfnGetDefaultContextCb_t)(
-    zer_get_default_context_params_t* params,
+    void* params,
```
joshuaranjan, line 2733: Why do we need void * params.
If empty params, if possible, lets skip this member variable

pratikbariintel: Currently all the callbacks defined have the same pattern. In case this field is not required we have to define a new callback pattern.

scripts/templates/tracing/trc_setters.h.mako
pratikbariintel: This added to have the same callback pattern as others defined.

joshuaranjan:
LGTM
Please request Vishnu to review

include/layers/zel_tracing_register_cb.h
vishnu-khanth, line 2733: The function pointer signature is changed in tracing header. 
Just thinking about whether this breaks backward compatibility

rwmcguir: So this data type is used in the exported function call: 

ZE_APIEXPORT ze_result_t ZE_APICALL
zelTracerGetDefaultContextRegisterCallback(
    zel_tracer_handle_t hTracer,
    zel_tracer_reg_t callback_type,
    zer_pfnGetDefaultContextCb_t pfnGetDefaultContextCb
    );
  
So while it won't break immediate compilation or exported function ABI, the public struct type zer_pfnGetDefaultContextCb_t is changing in it's struct members.  Thinking through this out loud.. The struct remains the same size, and void* also remains the same size. 

The problem is the callback itself, the user must typecast it to match our callback data type, i.e. you see this yourself in the test content. It breaks how the user must define the actual callback API. This is my fear that, yes, this breaks ABI compatibility.  Can instead of going this route, perhaps insert a placeholder void member; or something into the actual data structure, to avoid the clang warning?

joshuaranjan: >Can instead of going this route, perhaps insert a placeholder void member; or something into the actual data structure, to avoid the clang warning?

That would change the size of the structure ? Which might again be an ABI issue.

joshuaranjan: Is there a workable solution which would not break ABI.
The solutions we discussed seems to break ABI one way or the other.

@rwmcguir @pratikbariintel

pratikbariintel: If we consider the addition of a placeholder in the previous structure (zer_get_default_context_params_t), then this will also change its size. However, since this structure is not used anywhere inside the code, it might not be throwing the ABI issue.

If we go by this approach, I think it is better to name the placeholder something like 'reserved', so that the user doesn't think of using the same.

nrspruit:
This patch causes an ABI issue, which is not acceptable.

A cleaner method we chose to resolve this is https://github.com/oneapi-src/level-zero/pull/438 which will wrap these scenarios with a clang wrapper to ignore this issue.

In addition, a clang CI build has been added to catch this issue going forward.

rwmcguir:
Closing as we fixed this in another patch.

PR #445, opened 2026-04-21
documentation: Add readme on tests and loader versioning

nrspruit:


test/README.md
```diff
@@ -0,0 +1,365 @@
...
+- `driver_ordering_unit_tests.cpp` - Driver ordering logic (static builds and non-Windows only)
+- `driver_teardown_unit_tests.cpp` - Driver teardown logic (static builds and non-Windows only)
+- `init_driver_unit_tests.cpp` - Driver initialization tests (non-Windows only)
+- `init_driver_dynamic_unit_tests.cpp` - Dynamic driver initialization tests (dynamic builds, non-Windows only)
+
+### Build Configurations
+
+Different test files are conditionally compiled based on the build configuration:
+
+**Static Build Only (`BUILD_STATIC` or non-Windows):**
```
rwmcguir: So just confirming, this means these tests build for static loader and linux only?

test/README.md
```diff
@@ -0,0 +1,365 @@
...
+
+CTest automatically sets up the required environment variables, working directories, and driver paths needed for each test. Running the test executable directly (`./tests` or `./bin/tests`) will result in:
+- Missing environment configurations
+- Incorrect driver paths
+- Failed test assertions
+- Unreliable test results
+
+### Prerequisites: Set ZEL_LIBRARY_PATH
+
+**Before running CTest, you must set the `ZEL_LIBRARY_PATH` environment variable** to point to where the loader libraries and null drivers are located. This ensures tests can find the required libraries at runtime.
```
rwmcguir, line 60: Should we tell the user to not use LD_LIBRARY_PATH?  Or either? 
Or this is a better OS agnostic variable to be used?

test/README.md
nrspruit: These are only run on:
static builds on windows or linux
dynamic builds -> linux only

Due to how windows dirver discovery works.

test/README.md
nrspruit: We can say either, but to avoid altering other programs, ZEL_LIBRARY_PATH is specific to L0.

I prefer limiting this to ZEL_LIBRARY_PATH for the unit tests specifically.

rwmcguir (approved):

PR #448, opened 2026-04-27
Add Ubuntu specific .deb packages to CPack

rwmcguir:


.github/workflows/build-quick.yml
```diff
@@ -65,7 +65,7 @@ jobs:
           echo "✓ zello_world ran successfully with installed loader and NULL driver"
       - name: Uninstall loader
         working-directory: build
-        run: make uninstall || xargs rm -v < install_manifest.txt
+        run: make uninstall || xargs rm -v < install_manifest.txt || xargs rm -v < install_manifest.txt
```
rwmcguir: This needs fixed... We don't need duplicate

.github/docker/ubuntu.Dockerfile
```diff
@@ -52,6 +52,7 @@ apt-get install -y \
   libc6:arm64 \
   libstdc++6:arm64 \
   git \
-  ninja-build
+  ninja-build \
+  python3
```
rwmcguir, line 56: This removes a simple warning during the build as CMake still attempts to resolve this binary, even though not technically used in Linux.

nrspruit (approved):

PR #449, opened 2026-04-27
Fix typo in build debug command line

rwmcguir:
This should fix the remaining build errors in the ABI checker, we use a lot of lambda functions in the L0 Loader and there is specific command line option that needed enabled. Additionally disabling -lto ensure linker is not doing overly clever things that prevent ABI checkers in corner conditions.

### DIFF
--- a/.github/workflows/check-abi.yml
+++ b/.github/workflows/check-abi.yml
@@ -39,13 +39,13 @@ jobs:
       run: |
         mkdir base/build
         cd base/build
-        cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS="-Og -g" ..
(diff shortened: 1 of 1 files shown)


rwmcguir:
Note how clean this step is now:

Dump ABI Base: 
2m 38s
Run ./abi-dumper.pl \
  ./abi-dumper.pl \
    ./base/build/lib/libze_loader.so \
    -vnum $(cat ./base/build/VERSION) -lambda \
    -public-headers ./base/include \
    -o ./base.dump
  shell: /usr/bin/bash -e {0}
Detect public symbols
Reading debug-info
Reading v-tables
Creating ABI dump

The object ABI has been dumped to:
  ./base.dump

nrspruit (approved):

PR #461, opened 2026-05-11
Add install rule for zello_world sample binary

WithEnoughCoffee:
## Summary

The `zello_world` binary is built but never installed by CMake, which forces downstream packagers (e.g. Fedora) to manually copy it from the build directory using hard-coded paths like `redhat-linux-build/bin/zello_world`.

This adds an `install(TARGETS ...)` rule so that `cmake --install` places `zello_world` in the standard `${CMAKE_INSTALL_BINDIR}` directory, consistent with how other targets in this project are installed (e.g. `ze_loader`, `ze_tracing_layer`).

## Changes

- `samples/zello_world/CMakeLists.txt`: Add `install(TARGETS ${TARGET_NAME} RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})`

## Motivation

This eliminates the need for downstream packagers to reference build-directory internals. The Fedora RPM spec currently has:
```
install -p -m 755 ./redhat-linux-build/bin/zello_world %{buildroot}%{_bindir}/zello_world
```
With this change, `%cmake_install` would handle it automatically.

Signed-off-by: Autumn Nash <autumncapasso@gmail.com>


rwmcguir:
Yes, on purpose. Also, Fedora is not doing as you suggest:
https://packages.fedoraproject.org/pkgs/oneapi-level-zero/oneapi-level-zero/fedora-44.html#provides

zello_world is a sample binary and not part of the official installable package for a reason.  The shipping installable packages are for runtime use with drivers and applications.  zello_world is not useful this in this context. 

Large customers that use this in clusters do not want the file system polluted with sample applications that are unnecessary to use in production environments.

If a package were to distribute a simple sample application, then a third package would need to be generated such as level-zero-samples or oneapi-level-zero-samples (pending on which naming convention Fedora vs github) one follows. I personally don't see the ROI here, when people writing code are probably more interested in looking at the sample, as opposed to running it.

The major use case for zellow_world is two fold, 1 it is baked into our CI and CMakefile build process to ensure this library is indeed working and 2 it serves as sample source. Providing users a binary copy isn't super useful in most cases.

WithEnoughCoffee:
Ok Thanks.

PR #463, opened 2026-05-21
feature: add to_string support for callbacks

KamilKoprykIntel:
Related-To: NEO-14577
Signed-off-by: Kamil Kopryk <kamil.kopryk@intel.com>


rwmcguir (approved):
This look OK to me, we should also get feedback from @nrspruit

nrspruit (approved):

PR #473, opened 2026-06-10
New bb tests

aviralni:


bindings/sysman/python/source/cts_tests/test_power_cts.py
```diff
@@ -0,0 +1,388 @@
+##
```
shubskmr, line 1: move this to test/cts_tests/test_power_cts.py

bindings/sysman/python/source/cts_tests/test_power_cts.py
```diff
@@ -0,0 +1,388 @@
...
+                    device, byref(reduced_count), reduced_handles
+                )
+                assert (
+                    rc == self.pz.ZE_RESULT_SUCCESS
+                ), f"zesDeviceEnumPowerDomains(reduced) failed with ze_result_t={rc}"
+                assert len(reduced_handles) == reduced_count.value
+
+        assert found_power_support, "No power handles found on any of the devices"
+
+    def test_GivenSamePowerHandleWhenRequestingPowerPropertiesThenExpectSamePropertiesTwiceAndLimitsAreInRange(
```
shubskmr, line 200: Do we need such tests which seems to be similar to L0 CTS?
@joshuaranjan thoughts?

bindings/sysman/python/source/cts_tests/test_power_cts.py
```diff
@@ -0,0 +1,388 @@
...
+                    power_handle, byref(restore_count), restore_array
+                )
+                assert (
+                    restore_rc == self.pz.ZE_RESULT_SUCCESS
+                ), f"zesPowerSetLimitsExt(restore) failed with ze_result_t={restore_rc}"
+
+        if not exercised_any:
+            pytest.skip(
+                "No mutable root-device power limit descriptors were available for zesPowerSetLimitsExt"
+            )
```
shubskmr, line 388: in BMG I think we have 2 domains GPU and MEMORY which get the energy counters from PMT. So please add a test to check if L0 return insufficient permission error for these when test is not running as root.
@joshuaranjan KMD interface might be different for different platform for same APIs, what is the plan to scale these tests?
Today on BMG power APIs might return insufficient permission and the test might pass but what about next platforms?

shubskmr:
please use a better commit title like
test: Add new CTS tests in pyzes

rwmcguir:
As a request, your commit description needs to be a little more verbose. It is not obvious for a public repository what this is about or related to. I don't know for example what a bb test is as a maintainer of the primary L0 Loader product.

bindings/sysman/python/source/cts_tests/conftest.py
```diff
@@ -0,0 +1,74 @@
+##
+# Copyright (C) 2026 Intel Corporation
```
joshuaranjan, line 2: Move this under test/
instead of cts_tests folder name , please make it as blackbox_tests

Rename "[pyzes_black_box_test.py](https://github.com/oneapi-src/level-zero/pull/473/changes#diff-9fbeb6a435c18c420acbce35bbdb74f4c830d4f79f51b6914e5e34f77e718f77)" to pyzes_examples.py

bindings/sysman/python/source/cts_tests/conftest.py
```diff
@@ -0,0 +1,74 @@
...
+#
+##
+
+import os
+import sys
+from ctypes import byref, c_uint32
+
+import pytest
+
+CTS_TESTS_DIR = os.path.dirname(os.path.abspath(__file__))
```
joshuaranjan, line 14: rename all the variables.
Lets avoid CTS to avoid confusion with actual conformance tests

bindings/sysman/python/source/cts_tests/conftest.py
```diff
@@ -0,0 +1,74 @@
...
+    try:
+        import pyzes as pz
+    except (ImportError, OSError) as exc:
+        pytest.skip(f"pyzes is not available for CTS validation: {exc}")
+
+    return pz
+
+
+@pytest.fixture(scope="session")
+def sysman_devices(pyzes_module):
```
joshuaranjan, line 32: What is the meaning of "conftest" ?
Is it config tests ?

bindings/sysman/python/source/cts_tests/conftest.py
aviralni: ok

bindings/sysman/python/source/cts_tests/conftest.py
aviralni: conformance

joshuaranjan:
@aviralni 
Please share if any update here .. 
Comments seem to be very old

joshuaranjan:
Please move this to draft

PR #499, opened 2026-07-21
Customize log filename via pattern tokens in ZEL_LOADER_LOG_FILE

MaciejPlewka:


rwmcguir:
We have unit tests in the repo to test portions of this, perhaps you can add one to verify this operates?
You can ask AI to generate these, and put them as part of the CMakeFile in the ./test/ folder to execute with proper environment flags enabled.  The body of the code looks solid.

PR #502, opened 2026-07-31
POC for zelUnloadDrivers

rwmcguir:
POC Only for design discussion, not for merging yet.
May need to be moved to L0 Spec discssuion

At current drivers unloaded, cannot be reloaded due to initial implementation. This is the first major limitation in this design.  A design decision needs to be made here, if we support re-loading, or effectively treat this as a banned driver during this process execution.

Driver ordering is impacted, not re-ordering but a HOLE will develop in the schema, so if N devices loaded, and N-1 devices are left, it would not be clear to user which N's are valid. It appears very unsafe to attempt re-ordering after users may have driver handles, so list compaction seems like a bad idea. Returning UNITIALIZED seems best for now, or in the future UNLOADED is a potential upgrade. This specifically affects zer* API's if the first driver is unloaded, they will then be redirected to the next valid driver.

### DIFF
--- a/include/loader/ze_loader.h
+++ b/include/loader/ze_loader.h
@@ -171,6 +171,36 @@ zelLoaderTranslateHandle(
    void **handleOut);
 
 /**
+ * @brief [PROOF OF CONCEPT] Unloads a single Level Zero driver identified by its handle.
+ *
+ * This function unloads a driver that was previously reported by zeDriverGet()/zeInitDrivers().
+ * The driver's shared library is freed, its DDI tables are cleared, and the driver is removed
+ * from the loader's internal driver lists so it is no longer reported by subsequent enumeration.
+ *
+ * Preconditions / limitations (proof of concept):
+ * - The driver handle must correspond to a currently loaded driver.
+ * - The driver must be unused: all child objects created through the driver (contexts, command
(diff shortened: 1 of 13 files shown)


include/loader/ze_loader.h
```diff
@@ -170,6 +170,120 @@ zelLoaderTranslateHandle(
...
+ * @return
+ *   - ZE_RESULT_SUCCESS if the driver was successfully unloaded.
+ *   - ZE_RESULT_ERROR_INVALID_NULL_HANDLE if hDriver is NULL or does not match a loaded driver.
+ *   - ZE_RESULT_ERROR_UNSUPPORTED_FEATURE if hDriver is a driver DDI handle rather than a loader handle.
+ *   - ZE_RESULT_ERROR_HANDLE_OBJECT_IN_USE if the driver still owns live child objects and
+ *     ZEL_UNLOAD_DRIVER_FLAG_FORCE was not specified.
+ *   - ZE_RESULT_ERROR_UNINITIALIZED if the loader has not been initialized.
+ */
+ZE_APIEXPORT ze_result_t ZE_APICALL
+zelUnloadDriverExt(
```
rwmcguir, line 252: Just commentary I'll need to rename this if we keep it, probably wrap this into single API with all flags.. 
Ext while legal in the Loader, not something I want confusing people with normal Ext's from the spec.

PR #507, opened 2026-08-26
Fix Driver Ext Tracing Legacy driver compatability

nrspruit:


rwmcguir:
Two small things.

`pfnGet` is called twice, but the first call's count is only used for the `0 == driverCount` check — which the second call already reports. One call would do.

Bigger question: the loader keys extension registration and fan-out per handle (`ExtensionFunctionKey{hDriver, functionName}` in `tracing_imp.h`), but this gates per library using one cached handle. That's fine if `zelDriverEnableTracing` is defined as library-global — all handles in a driver share a code base, so one handle can speak for all of them. Could the header say that explicitly? Otherwise a driver that keeps the flag per handle silently traces only the first one.

rwmcguir (approved):

PR #511, opened 2026-09-09
fix: missing ZET_METRIC_SOURCE_ID_EXP_NAME

matcabral:


rwmcguir:
This looks like something that should be posed into the spec itself, in the file level-zero-spec/scripts/tools/metricGroupMarker.yml file. We can then absorb that change when the loader picks up Spec 1.19 fairly soon.

PR #516, opened 2026-09-16
Update L0 Loader w/L0 Spec 1.19.12

rwmcguir:
+ Update CHANGELOG.md for Spec and New features.


rwmcguir:
Verified the ABI report is throwing false alarms for two issues, will be merging this.

matcabral:
Tools headers +1.

Jemale (commented):
+1, the abi check is weird like you said. Do we think it's just noise?

rwmcguir:
It is, I've actually fix that in the Spec as we run it every PR, but here it only runs once in a while and it was addressed already in the Spec Repository. I've dug into the details of each and we are clear (at least on the two being flagged)

